PR #82, opened 2015-03-23
Blacklist workers which cannot execute parrot.

### LINKED ISSUE #78: kill problematic workers
Certain types of errors are caused by issues with the worker machine's configuration (see error 127 [here](http://www.crc.nd.edu/~awoodard/lobster/scale_vs_overhead_aod_1_core_workers_10_events_per_job_v21/) for example, which we think is caused because by workers landing on RH5 machines.) We already keep track of the hostname and exit code. We should intelligently kill workers that only produce failed tasks with the same exit code. @dthain, @btovar: I don't see a way to blacklist a particular worker [here](http://ccl.cse.nd.edu/software/manuals/api/html/classwork__queue_1_1WorkQueue.html), but I might be missing it. Is this functionality something you'd consider adding to Work Queue?

matz-e:
This is work in progress and needs more testing.  Creating PR to not forget about it…

Fixes #78.

### DIFF
--- a/doc/ErrorCodes.md
+++ b/doc/ErrorCodes.md
@@ -4,6 +4,7 @@ Lobster uses the following error codes:
 
 | Code | Reason
 | :--- | :-----
+|  169 | Unable to run parrot
 |  170 | Sandbox unpacking failure
 |  171 | Failed to determine base release
 |  172 | Failed to find old releasetop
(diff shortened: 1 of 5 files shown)


annawoodard:
I have also tested this. It is great. It completely solves my mysterious madgraph compilation issue, where I can't figure out what it is about certain nodes that causes my jobs to fail.

I vote that we merge the PR now, and open any improvements we want to make (handling transient errors, better feedback to the user about permanent errors and how they should update their submission) in a new issue.

matz-e:
OK.

Issue #85, opened 2015-03-30
Refactor non-cmssw-specific bits out of lobster.cmssw.JobProvider up to the parent class


annawoodard:
We currently have four job provider configurations: 

1. lobster.job.SimpleJobProvider
2. lobster.job.JobProvider
3. lobster.cmssw.JobProvider for cmsRun jobs
4. lobster.cmssw.JobProvider for non-cmsRun jobs.

While a little confusing, the categorization makes sense: lobster.cmssw.JobProvider is not for cmsRun jobs specifically; it is for jobs for which we need the cmssw environment setup before executing. As things have evolved, a lot of stuff that is not specific to type 3 jobs above has ended up there-- most importantly, the bits having to do with file transfer in and out (currently we only have chirp and xrootd capabilities for type 3 jobs.) We should consider refactoring these bits up to the parent class.

klannon:
@annawoodard: Thanks for writing an issue for this.  I've currently assigned it to @khurtado, not because I doubt your or Matthias' ability to handle this, but because I believe that both you and Matthias have other fish to fry.  @khurtado: If its not clear what needs to be done here, then lets talk (maybe after the Thursday meeting?

khurtado:
@klannon sounds good! I understand what the issue is and what needs to be done, but I'm sure I will have questions once I start looking and working on those bits...

matz-e:
@annawoodard after our discussion yesterday, I think that maybe we should remove the the `SimpleJobProvider` and count on everything using the database?

Then the `JobHandler` could be split into one that deals with simple configurations, and a `FrameworkJobHandler`.  The wrapper would have to be "abstracted" in a similar fashion.  Then we should change the database to not have columns of run, lumi but some more generic "id" that expands to a python tuple of sorts.

Currently, I'd have some interest in getting "pythia"-style MC generation going, since I can't run this with crab ATM, which could drive this in a way…

Maybe we should have a short sync on the strategy with this?

klannon:
Based on our discussion this morning, I'm trying to indicate that this issue should be addressed before we make Lobster available to non-expert/non-ND users.  I marked the milestone as "Version 1 final" which may be hijacking the original intention of that milestone.  If you think we should have a different milestone to indicate this, that's fine too (although I'm not sure how to make new milestones...)

matz-e:
OK.  I've changed the due date of said milestone, too.

matz-e:
[...]

PR #87, opened 2015-04-01
Implement native HDFS functions via the hadoopy module. Fixes #86.

### LINKED ISSUE #86: Lobster breaks if the hadoop mount point detaches and chirp is not enabled.
We are currently reliant on fuse for all master-HDFS interactions. Fuse is finicky. We should implement native HDFS functions.

annawoodard:
This is not a very elegant solution-- I don't know how to detect HDFS endpoints except testing for a leading ‘hadoop' and making sure that all lobster.fs functions preserve the leading ‘hadoop.' But it works, so I think we should keep it as a starting point. I will merge this PR tomorrow unless there are any objections.

### DIFF
--- a/lobster/cmssw/dataset.py
+++ b/lobster/cmssw/dataset.py
@@ -1,9 +1,8 @@
 from collections import defaultdict
-import glob
 import math
 import os
 import sys
-from lobster import util
+from lobster import util, fs
 
 sys.path.insert(0, '/cvmfs/cms.cern.ch/crab/CRAB_2_10_2_patch2/external/dbs3client')
 from dbs.apis.dbsClient import DbsApi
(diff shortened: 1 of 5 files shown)


matz-e:
I like the implementation!  But… there's some functionality already in place to use chirp to check for files when merging etc, see [here](https://github.com/matz-e/lobster/blob/master/lobster/chirp.py) for the implementation and [here](https://github.com/matz-e/lobster/blob/master/lobster/cmssw/job.py#L334) for usage.

I would suggest the following ([before merging, as we should switch to a different development style, IMO](http://nvie.com/posts/a-successful-git-branching-model/)):

* use your module and decorator structure
* add  chirp functionality
* replace chirp module usage with the fs module

I would like to retain chirp usage, as the storage element is not guaranteed to be a local hadoop.  On the other hand, using your module will circumvent another layer of chirp to go through.  In the long term, we maybe should add some SRM capabilities, to not rely on chirp too much.  I think I had mentioned this to Kevin already, too.

annawoodard:
> I like the implementation!  But… there's some functionality already in place to use chirp to check for files when merging etc, see [here](https://github.com/matz-e/lobster/blob/master/lobster/chirp.py) for the implementation and [here](https://github.com/matz-e/lobster/blob/master/lobster/cmssw/job.py#L334) for usage.

Of course I know about the chirp module :). It's very nice, but in my last workflow I could not use it because I was not running a cmsRun executable, so I could not use Chirp (see #85). I've edited #86 for clarity.

> I would suggest the following ([before merging, as we should switch to a different development style, IMO](http://nvie.com/posts/a-successful-git-branching-model/)):

I agree-- let's switch to this development style.

> * use your module and decorator structure
* add  chirp functionality
* replace chirp module usage with the fs module

> I would like to retain chirp usage, as the storage element is not guaranteed to be a local hadoop.  On the other hand, using your module will circumvent another layer of chirp to go through.  In the long term, we maybe should add some SRM capabilities, to not rely on chirp too much.  I think I had mentioned this to Kevin already, too.

Sounds good to me.

matz-e:
The chirp module is used by the JobProvider, so your (non)usage of `cmsRun` does not really matter… None of the chirp module is used on the job.  That's why I'm proposing it should be folded into your new fs module.  Then we should be having several access modes for the data storage (ordered by priority):

* direct via the os module
* via hadoop module
* via chirp
* via srm (in the future)

annawoodard:
But the chirp module won't help you if you don't have a chirp server going.

matz-e:
That's why I'm suggesting folding it into the `fs` module to have all the options :)

The chirp module had a fall-back to local methods, so it didn't really require a running chirp server.

annawoodard:
But the local methods would fail if the storage element was in HDFS. I was just clarifying why I disagreed with the statement that the 'usage of cmsRun does not matter.'

annawoodard:
As per @matz-e's suggestion, I am closing this PR and re-opening it to merge with the dev branch. I'll open a new issue for @matz-e's suggestion to merge the fs and chirp modules.

annawoodard:
Since we have #91 opened now, and we changed the development model back to merging with the master, I'm going to merge this (again.)

PR #94, opened 2015-04-15
Add option to force processing, even if the working directory is locked by a previous instance.

### LINKED ISSUE #93: Add a --force option for the 'lobster process' command.
Until we address #83, it would be convenient to add a force option for the 'lobster process' command which would remove lobster.pid.

annawoodard:
Fixes #93. @matz-e, does this look ok?


annawoodard:
Done.

Issue #101, opened 2015-05-08
Support remote SE via SRM


matz-e:
The master still relies heavily on direct/hadoop/chirp access to the SE.  We should be able to manage remote storage with SRM, too.

matz-e:
Some notes on GFAL2:

```
export GFAL_CONFIG_DIR=/cvmfs/oasis.opensciencegrid.org/mis/osg-wn-client/3.2/3.2.22/el6-x86_64/etc/gfal2.d/
export GFAL_PLUGIN_DIR=/cvmfs/oasis.opensciencegrid.org/mis/osg-wn-client/3.2/3.2.22/el6-x86_64/usr/lib64/gfal2-plugins/
source /cvmfs/oasis.opensciencegrid.org/osg-software/osg-wn-client/3.2/current/el6-x86_64/setup.sh
```
Then
```
import gfal2
```
in python.

klannon:
Is this a replacement for the SRM utilities?

matz-e:
It's basically an abstraction, with SRM being one plugin.  I haven't found any python bindings for SRM so far, and I do not want to parse the SRM output by hand.  `gfal-copy` for example, can handle SRM URLs.

The [website](https://dmc.web.cern.ch/projects/gfal-2/home) has a list of protocols that are supported… and it is present in the same CVMFS repository as SRM.

matz-e:
Expanding on this: [here](https://github.com/matz-e/lobster/commit/447724805dda8da269c983b14594c49b0d3a7a00) I've removed almost all dependencies on the chirp module (save for checking that a chirp server is running).

I think we should abstract this out into a `se/chirp.py` and `se/cms.py` module, where the latter uses xrootd for data streaming and SRM for stage-out and managing.  For merge job creation, this may be a little more involved, as we might incur a penalty for talking to a remote server for some operations.  This may lead to some operations being async or the whole merge job creation/checking being done in a separate process.

These SE modules then would have to be hooked up to the `fs` module to opaquely switch how we're dealing with the SE?  `fs` is currently used in the `core`, `job`, `cmssw/dataset`, `cmssw.job` modules.

annawoodard:
Hi @matz-e. I think this all sounds good, but I think that the issues brought up during testing of #106 should be addressed first. I'll update there from the email thread.

matz-e:
Progress!

PR #106, opened 2015-05-12
Implement SRM stage-out for tasks.

### LINKED ISSUE #98: Stage-out via SRM/gridftp
We need another stageout mode to operate without Chirp and the penalty of transferring data via WQ.

matz-e:
This currently only allows the tasks to stage their files out via SRM.  The
master has no SRM connection to the storage element.  Fixes #98.

Uses either gfal-copy, if available, or lcg-cp, which is deprecated,
but more widespread in use.

@annawoodard any comments?  Do you want to test this?
@klannon this may need some fine-tuning for cms-connect, depending on the site configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,25 @@ The last line of arguments corresponds to the desired worker configuration.
 
 # Advanced usage
 
-## Stage-out via chirp
+## Stage-out
+
+### Via SRM
+
+Lobster supports SRM stage-out.  Parameters for SRM stage-out can be found
+in `/cvmfs/cms.cern.ch/SITECONFIG`, the directory of the destination
+storage element, file `PhEDEx/storage.xml`.  A `lfn-to-pfn` leaf with the
+`srmv2` protocol will reveal the server URL, which should be used in the
+configuration, without any `$1`.  For example, the adjusted settings for
+`T3_US_ND` are:
+
+    srm url: "srm://ndcms.crc.nd.edu:8443/srm/v2/server?SFN=/hadoop/"
+    srm root: "/hadoop/"
+
+The `srm root` maps the physical file name to the logical file name.  Both
+the `srm url` and `srm root` should be adjusted such that the logical file
+name can be appended to the `srm url` without any problem.
+
+### Via chirp
 
 Using chirp for stage-in and stage-out can be helpful when standard CMS
 tools for file handling, i.e., XrootD and SRM, are not available.
--- a/lobster/cmssw/data/job.py
+++ b/lobster/cmssw/data/job.py
@@ -222,10 +222,10 @@ def copy_outputs(data, config, env):
 
             prg = []
 
-            if int(os.environ["LOBSTER_GFAL_COPY"]) == 1:
-                prg = ["env", "-i", "X509_USER_PROXY=proxy", "gfal-copy"]
-            elif int(os.environ["LOBSTER_LCG_CP"]) == 1:
-                prg = ["lcg-cp", "-b", "-v", "-D", "srmv2"]
+            if len(os.environ["LOBSTER_GFAL_COPY"]) > 0:
+                prg = ["env", "-i", "X509_USER_PROXY=proxy", os.environ["LOBSTER_GFAL_COPY"]]
+            elif len(os.environ["LOBSTER_LCG_CP"]) > 0:
+                prg = [os.environ["LOBSTER_LCG_CP"], "-b", "-v", "-D", "srmv2"]
             else:
                 raise RuntimeError("no stage-out method available")
 
@@ -238,9 +238,10 @@ def copy_outputs(data, config, env):
             print " ".join(args)
             print "---"
 
+            # FIXME is this really needed after simplifying the repos?
             pruned_env = dict(env)
-            for k in ['LD_LIBRARY_PATH', 'PATH']:
-                pruned_env[k] = ':'.join([x for x in os.environ[k].split(':') if 'CMSSW' not in x])
+            # for k in ['LD_LIBRARY_PATH', 'PATH']:
+            #     pruned_env[k] = ':'.join([x for x in os.environ[k].split(':') if 'CMSSW' not in x])
 
             p = subprocess.Popen(args, env=pruned_env, stderr=subprocess.PIPE)
             p.wait()
--- a/lobster/cmssw/data/job.py
+++ b/lobster/cmssw/data/job.py
@@ -248,6 +248,8 @@ def copy_outputs(data, config, env):
             if p.returncode != 0:
                 data['stageout exit code'] = p.returncode
                 raise IOError("Failed to transfer output file '{0}':\n{1}".format(localname, p.stderr.read()))
+            else:
+                print p.stderr.read()
         elif chirp_server:
             if chirp_root and remotename.startswith(chirp_root):
                 remotename = remotename.replace(chirp_root, '', 1)
--- a/lobster/cmssw/data/job.py
+++ b/lobster/cmssw/data/job.py
@@ -236,10 +236,11 @@ def copy_outputs(data, config, env):
             print " ".join(args)
             print "---"
 
-            status = subprocess.call(args, env=env)
-            if status != 0:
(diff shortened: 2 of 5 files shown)


annawoodard:
Looks good to me. I will test, then merge.

annawoodard:
On testing, I saw the following error:

    lcg-cp: /cvmfs/cms.cern.ch/slc5_amd64_gcc462/cms/cmssw/CMSSW_5_2_5/external/slc5_amd64_gcc462/lib/libz.so.1: version `UUID_1.0' not found (required by /cvmfs/oasis.opensciencegrid.org/mis/osg-wn-client/3.2/3.2.23/el6-x86_64/usr/lib64/libgfal.so.1)

I suspect this is because I was using a different CMSSW version than you. Taking any CMSSW paths out of ```PATH``` and ```LD_LIBRARY_PATH``` solves the issue for me (I've pushed up a quick hack in 745f080, please improve it if you see a better solution.)

One other (minor) issue: if the user has both a chirp server and an SRM url defined, check_outputs() seems to fail and the job will return 211. I've just removed my chirp server for now, but I like the idea of being able to have both defined, in order to 1) have a backup if SRM fails and 2) check the output, if the user wants to, until we have a way to do so with lcg-cp/gfal-copy. I haven't looked into it to figure out why it's failing-- I'm pretty sure it is not an issue with my chirp server (because I was running successful jobs using it just before testing this), but that is always a possibility.

annawoodard:
I get the following error when submitting to the T3. @matz-e, do you think we should add a check in ```copy_outputs()``` on whether the SE is already accessible? 

    --- staging-out with:
    lcg-cp -b -D srmv2 file:////var/condor/execute/dir_7778/worker-174873-7821/t.1612/Hadronizer_MgmMatchTuneZ2star_8TeV_madgraph_cff_py_GEN_SIM_DIGI_L1_DIGI2RAW_HLT_RAW2DIGI_L1Reco_RECO_PU.root srm://ndcms.crc.nd.edu:8443/srm/v2/server?SFN=/hadoop/store/user/awoodard/LHE_v13/merge_a_b_d_e/AOD_v1/ttH/AHVMWOJMWK/Hadronizer_MgmMatchTuneZ2star_8TeV_madgraph_cff_py_GEN_SIM_DIGI_L1_DIGI2RAW_HLT_RAW2DIGI_L1Reco_RECO_PU_6367.root
    ---
      File "job.py", line 64, in check_execution
        yield
      File "job.py", line 508, in <module>
        copy_outputs(data, config, env)
      File "job.py", line 248, in copy_outputs
        raise IOError("Failed to transfer output file '{0}':\n{1}".format(localname, p.stderr.read()))
    IOError: Failed to transfer output file 'Hadronizer_MgmMatchTuneZ2star_8TeV_madgraph_cff_py_GEN_SIM_DIGI_L1_DIGI2RAW_HLT_RAW2DIGI_L1Reco_RECO_PU.root':
    [SE][Mkdir][] httpg://ndcms.crc.nd.edu:8443/srm/v2/server: CGSI-gSOAP running on d6copt314.crc.nd.edu reports Error reading token data header: Connection closed
    Internal error
    lcg_cp: Communication error on send

matz-e:
> I suspect this is because I was using a different CMSSW version than you. Taking any CMSSW paths out of PATH and LD_LIBRARY_PATH solves the issue for me (I've pushed up a quick hack in 745f080, please improve it if you see a better solution.)

This seems good.  I'd maybe change it to `pruned_env = dict(env)`, to avoid changing `env` in case it will be re-used later?

Also note that `gfal-copy` is called via `env -i`, which ignores the environment.  Maybe the same trick works with `lcg-cp`?  Probably not, though, due to custom paths.

>  One other (minor) issue: if the user has both a chirp server and an SRM url defined, check_outputs() seems to fail and the job will return 211. I've just removed my chirp server for now, but I like the idea of being able to have both defined, in order to 1) have a backup if SRM fails and 2) check the output, if the user wants to, until we have a way to do so with lcg-cp/gfal-copy. I haven't looked into it to figure out why it's failing-- I'm pretty sure it is not an issue with my chirp server (because I was running successful jobs using it just before testing this), but that is always a possibility.

Hm… I would have thought that that should work.  We should make it work.  I also like the notion of chirp as a fallback.

> I get the following error when submitting to the T3. @matz-e, do you think we should add a check in copy_outputs() on whether the SE is already accessible? 

Seems a little late for a check, then?  I'd maybe prefer having the chirp fallback.  But it looks like that error needs to be caught better!

annawoodard:
I'm not currently able to use SRM stageout at all. It all fails with the same 'communication on send' error I was getting on the T3 above. Trying a test run doesn't work, either:

    [earth] ~/lobster-dev >lcg-cp -b -v -D srmv2 file:////afs/crc.nd.edu/user/a/awoodard/test.txt        srm://ndcms.crc.nd.edu:8443/srm/v2/server?SFN=/hadoop/store/user/awoodard/lobster/spam
    Using grid catalog type: UNKNOWN
    Using grid catalog : (null)
    VO name: cms
    Checksum type: None
    Destination SE type: SRMv2
    Destination SRM Request Token: put:43
    Source URL: file:/afs/crc.nd.edu/user/a/awoodard/test.txt
    File size: 6
    Source URL for copy: file:/afs/crc.nd.edu/user/a/awoodard/test.txt
    Destination URL: gsiftp://ndcms.crc.nd.edu//hadoop/store/user/awoodard/lobster/spam
    # streams: 1
            0 bytes      0.00 KB/sec avg      0.00 KB/sec inst
            0 bytes      0.00 KB/sec avg      0.00 KB/sec inst
            0 bytes      0.00 KB/sec avg      0.00 KB/sec inst
            0 bytes      0.00 KB/sec avg      0.00 KB/sec inst
            0 bytes      0.00 KB/sec avg      0.00 KB/sec inst
            0 bytes      0.00 KB/sec avg      0.00 KB/sec inst
            0 bytes      0.00 KB/sec avg      0.00 KB/sec inst
            0 bytes      0.00 KB/sec avg      0.00 KB/sec inst
            0 bytes      0.00 KB/sec avg      0.00 KB/sec inst
            0 bytes      0.00 KB/sec avg      0.00 KB/sec inst
            0 bytes      0.00 KB/sec avg      0.00 KB/sec inst
            0 bytes      0.00 KB/sec avg      0.00 KB/sec inst
    file:/afs/crc.nd.edu/user/a/awoodard/test.txt: zero number of replicas
    lcg_cp: Invalid argument

I'm not sure how to debug. Any ideas?

annawoodard:
Also, as discussed in chat, we should add a checkwrite before running when SRM stageout is enabled.

annawoodard:
It looks like this is a permissions issue, I'll verify later today.

matz-e:
If we implement #101, we get a checkwrite equivalent for free, as we create the project output directories…

annawoodard:
@matz-e, I implemented your suggestion about pruned_env and added a check on whether the stageout directory is already accessible-- can you review it? I haven't checked yet the status of the permissions issue, I figure it's best to back up and test with only srm_stageout, which is giving this error:

    ----- Begin Fatal Exception 20-May-2015 10:59:31 EDT-----------------------
    An exception of category 'Incomplete configuration' occurred while
       [0] Constructing the EventProcessor
       [1] Constructing ESSource: class=PoolDBESSource label='GlobalTag'
    Exception Message: 
    Valid site-local-config not found at /cvmfs/cms.cern.ch/SITECONF/local/JobConfig/site-local-config.xml
    ----- End Fatal Exception -------------------------------------------------

Once we're done iterating with @btovar on that in the email thread, I'll report on the status of the permissions issue.

matz-e:
Looks good.  Maybe we should print a message for local stage-out?

I'll test this with Ben's suggestions later.

matz-e:
@annawoodard I think this branch now is getting too much into fixing smaller issues with CVMFS.  I think I should add a copy check and then we could merge this into the master?

annawoodard:
Sure, but can we open the remaining CVMFS issues in the tracker so we don't forget them?

PR #107, opened 2015-05-12
 Rework job cancellation.

### LINKED ISSUE #105: Speed up `lobster terminate …`
The termination command of lobster should be fast.  Currently it supplies the dashboard with a lot of updates, which is quite slow.  We should do the following:

* move the dashboard updates to the main loop,  where we actually test if we should quit
* print an INFO statement so that the user knows what is going on

matz-e:
Upon lobster termination, call a method in the job source, which for CMS
jobs sets the dashboard status to 'killed'.  They then show up in the
dashboard page as 'Cancelled' instead of 'Failed'.

Removes duplicated job reporting and moves it from `core.py` to the JobProvider.

Fixes #105.

### DIFF
--- a/lobster/cmssw/dash.py
+++ b/lobster/cmssw/dash.py
@@ -17,6 +17,7 @@ SUBMITTED = 'Pending'
 DONE = 'Done'
 RETRIEVED = 'Retrieved'
 ABORTED = 'Aborted'
+CANCELLED = 'Killed'
 RUNNING = 'Running'
 WAITING_RETRIEVAL = 'Waiting Retrieval'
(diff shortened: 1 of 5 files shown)


annawoodard:
AWESOME

Minor thought: why not use consistent language with Dashboard? (i.e., 'killed' instead of 'cancelled')

matz-e:
That's what CRAB 3 does in the DagmanKiller.  The dashboard then translates it to 'CANCELLED'.

annawoodard:
I see. Thanks for the clarification!

PR #111, opened 2015-05-14
Core dumps

matz-e:
Opening this for reference, will update later.  These commits will give you `core.*` files when the WQ bindings crash (or so I hope, works killing lobster with `-SEGV`.)


annawoodard:
@matz-e, this looks good to me, can you resolve the conflicts and merge?

PR #126, opened 2015-06-01
Add a generic pset for the example configuration.

annawoodard:
Hi @matz-e, I threw this together in a hurry, because I wanted something Allie could run out-of-the-box. Can you take a look? Particularly in [slim.yaml](https://github.com/matz-e/lobster/blob/16d3223f157935fe571c7db54e483985a5f37d8c/examples/slim.yaml#L51), where I tried to clarify the task options somewhat, as they are getting quite unweildy.


matz-e:
I've only adjusted minor things in there.  I think you've summarized everything pretty well.

One point that keeps tripping me up is our verbally agreed nomenclature clashes with our config file naming scheme.  As in "a configuration file (workflow) contains several jobs (currently named tasks) that are broken down into tasks"…

annawoodard:
Thanks for adjusting things, @matze! I agree about the nomenclature, I have a few things to add so I'm opening it in a new issue.

PR #128, opened 2015-06-02
Remove the Parrot linker hack.

annawoodard:
Hi @matz-e, I've tested that this works for me. We should test with your workflow too, so I'm opening this PR rather than committing to the master.


matz-e:
Works for me, too.  Tested ntuplizing on the opportunistic queue.

PR #130, opened 2015-06-03
Add non-wq stageout/stagein for non-cmsRun jobs.

annawoodard:
This is a quick hack, but I don't think it makes the issues in #85 worse. I'd like to merge it as an intermediate step before starting on the implementation of #85 so that we can use non-wq stageout/stagein for non-cmsRun jobs in the meantime.


matz-e:
Honestly, this seems like 90% of the work already done.  I'll have another look at it later today.  One thing I'd like to evaluate is the clauses for the dashboard reporting:  most of the if'ed out (in case of non-cmsRun) statements could actually also be provided for non-cmsRun jobs/tasks.

On the job reporting:  I've recently thought it might be better to parse a report when present, and if not instead fill the report hash/json we pass back to master.
That way, e.g., if I write a `hadd` merging setup, the custom script can provide a report, and not have one generated.  Maybe I'm overlooking something, though.

PR #131, opened 2015-06-03
Remove CMSSW-specific bits from env before srm stageout.

### LINKED ISSUE #129: lcg-cp is hardcoded for slc6
We currently have the OSG setup [hard-coded](https://github.com/matz-e/lobster/blob/master/lobster/cmssw/data/wrapper.sh#L123) for slc6. This causes srm stageout to fail if the user is running cmssw compiled on slc5. Switching `el6-$(uname -m)` to `el5-$(uname -m)` solves the problem. Do we want to figure out a way to determine the correct setup script based on the cmssw version rather than the machine? It might also make sense to say 'slc5 code is not supported, if you want to run crappy old code it's your own responsibility to hack the wrapper'.

    /cvmfs/oasis.opensciencegrid.org/mis/osg-wn-client/3.2/3.2.23/el6-x86_64/usr/bin/lcg-cp: /cvmfs/cms.cern.ch/slc5_amd64_gcc462/cms/cmssw/CMSSW_5_2_5/external/slc5_amd64_gcc462/lib/libuuid.so.1: version `UUID_1.0' not found (required by /cvmfs/oasis.opensciencegrid.org/mis/osg-wn-client/3.2/3.2.23/el6-x86_64/usr/lib64/libgfal.so.1)

annawoodard:
This is necessary in case the user is running CMSSW compiled on
a different OS than the host. Also, parse the current SLC version
rather than hardcoding it. Fixes #129.

### DIFF
--- a/lobster/cmssw/data/job.py
+++ b/lobster/cmssw/data/job.py
@@ -241,10 +241,9 @@ def copy_outputs(data, config, env):
             print " ".join(args)
             print "---"
 
-            # FIXME is this really needed after simplifying the repos?
             pruned_env = dict(env)
-            # for k in ['LD_LIBRARY_PATH', 'PATH']:
-            #     pruned_env[k] = ':'.join([x for x in os.environ[k].split(':') if 'CMSSW' not in x])
+            for k in ['LD_LIBRARY_PATH', 'PATH']:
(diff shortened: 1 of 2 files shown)


matz-e:
Point taken.  On a side note, in CRAB, `cmsRun` is run in a dedicated environment, while the wrapper and stage-out run in the node default environment (for future reference.)

PR #135, opened 2015-06-13
Use any SE for stageout

matz-e:
Tested at ND with:
* WQ
* Chirp
* XrootD/SRM from/to CERN

At CERN with:
* XrootD/SRM

Still missing:
* [x] test for availability of chirp server / lcg-tools
* [x] fix data discovery
* [x] complete chirp support
* [x] automatic siteconfig parsing for URLs


lobster/se.py
```diff
@@ -0,0 +1,313 @@
...
+        self.remove = hadoopy.rmr
+
+        # local imports are not available after the module hack at the end
+        # of the file
+        self.__hadoop = hadoopy
+
+    def isfile(self, path):
+        return self.__hadoop.stat(path, '%F') == 'regular file'
+
+class Chirp(FileSystem):
```
annawoodard: I think Chirp has python bindings now. Since we already have this functionality here, I don't think it's a big priority to get them integrated, but let's keep it in mind.

matz-e:
Have you tried using the bindings?  They fail with some globus library confusion for me.  I'm not sure how solvable that is…

README.md
```diff
@@ -155,24 +155,33 @@ The last line of arguments corresponds to the desired worker configuration.
...
+Lobster supports multiple file transfer methods:  via Work Queue, Chirp,
+XrootD (input only), and SRM (output only.)  Basic settings are:
 
-Lobster supports SRM stage-out.  Parameters for SRM stage-out can be found
-in `/cvmfs/cms.cern.ch/SITECONFIG`, the directory of the destination
-storage element, file `PhEDEx/storage.xml`.  A `lfn-to-pfn` leaf with the
-`srmv2` protocol will reveal the server URL, which should be used in the
-configuration, without any `$1`.  For example, the adjusted settings for
-`T3_US_ND` are:
+    storage:
```
annawoodard: I think we need to clarify what 'base' means here.

matz-e: Yes.  Depending on how all the other paths are defined (they should all point to the same directory), it could be the LFN in the CMS definition, but also the PFN.  Nothing should stop you from using `/hadoop/store/user/…`

matz-e:
Currently also testing this at CERN.  One more action item:

- [ ] test for availability of chirp server / lcg-tools

lobster/fs.py
```diff
@@ -1,55 +1,9 @@
...
-    res = glob.glob(path)
-
-    if len(res) == 0:
-        try:
-            res = ['/hadoop' + x for x in hadoopy.ls(path.replace('/hadoop', '', 1))]
-        except:
-            raise
-
-    return res
+from se import FileSystem, Local, Hadoop, SRM
```
annawoodard: I'm not sure I follow the motivation for having separate fs and se modules now (sorry if I'm missing something obvious). After all, the base class in se is called FileSystem.

matz-e: It's more of a technical nature.  The trick pulled off in the `fs` module to keep the direct replacement of the `os.path` methods didn't allow to import the classes in the `fs` module, so I stuck them into the `se` module, as they tend to access storage elements.  Maybe the `FileSystem` class is a bit of a misnomer?

matz-e: Could also call that class `StorageElement`…

annawoodard: >  Could also call that class `StorageElement`…

I like that idea. Especially since `SRM` is a FileSystem class but SRM itself isn't a particular filesystem. But you already have a `StorageElement` class. Maybe the `StorageElement` class could be some kind of Helper or Handler or something?

annawoodard:
AWESOME

PR #152, opened 2015-07-23
Handle task results from WQ. Fixes #80.

### LINKED ISSUE #80: Check returning tasks more thoroughly for failures
WorkQueue provides flags which indicate a task status (see [docs](http://ccl.cse.nd.edu/software/manuals/api/html/classwork__queue_1_1Task.html#a96170c2cb7416e66882d83c1b2d25b18)).  We should check if files are missing, resources exceeded, etc.

annawoodard:
I believe this is the correct subset of WQ results for which we
want the WQ result to trump whatever exit code we're returning. For
missing files (wq.WORK_QUEUE_RESULT_{INPUT|OUTPUT}_MISSING, let's
handle it on our side (trust the wrapper to return a sensible exit
code), as it gives us more control.

### DIFF
--- a/doc/ErrorCodes.md
+++ b/doc/ErrorCodes.md
@@ -4,6 +4,10 @@ Lobster uses the following error codes:
 
 | Code | Reason
 | :--- | :-----
+|  4   | The task ran but its stdout has been truncated
+|  8   | The task was terminated with a signal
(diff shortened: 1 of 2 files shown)


matz-e:
Sweet!

PR #159, opened 2015-07-28
Change hot cache behavior.

### LINKED ISSUE #157: Change cache definition behavior
Write the cache file with a timestamp immediately after setting the cold cache flag, and have all subsequent job check that their time is greater than the saved one to be hot.

matz-e:
Moves the hot cache check to compare against a file with a timestamp:
The first finished job records its time of the processing of the first
event (CMS jobs) or its end time (all others).  Consecutive jobs compare
their startup time to said time, and report themselves as "hot" when their
wrapper start time lies beyond the cache flag time.

Fixes #157.


annawoodard:
Awesome!

PR #160, opened 2015-07-28
Create a SiteDBJSON for all monitor instances. Fixes #112.

### LINKED ISSUE #112: Log issues when dashboard is set to false
It is useful to disable the dashboard when doing testing (although, as I've stated elsewhere, we should return to having the default setting be true.) Currently if I disable the dashboard, the log hangs. Has anyone else observed this?

annawoodard:
I have NO IDEA why this works but I've verified carefully that it does. @khurtado, @matz-e, can anyone offer any insight? I'm sure this is the wrong way to solve the problem since the whole idea of a dummy monitor is that it shouldn't be dependent on CMSSW-specific bits. On the other hand, I'd like to be able to disable monitoring in order to avoid polluting our dashboard records with bogus test runs, so unless someone can improve this fix, I vote we 'merge now, ask questions later.'

### DIFF
--- a/lobster/cmssw/dash.py
+++ b/lobster/cmssw/dash.py
@@ -37,6 +37,7 @@ status_map = {
 class DummyMonitor(object):
     def __init__(self, workdir):
         self._taskid = util.checkpoint(workdir, 'id')
+        self._db = SiteDBJSON({'cacheduration': 24})
 
     def generate_ids(self, jobid):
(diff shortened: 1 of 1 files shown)


annawoodard:
Clearly it has something to do with the multiprocessing and/or daemon modules.

khurtado:
+1 for merging and investigating further later

PR #163, opened 2015-07-29
Use task times parsed from non-cmsRun jobs.

annawoodard:
For the three cmssw-specific times, use time_prologue_end as a default.


matz-e:
Looks good, thanks!

PR #164, opened 2015-07-30
Se tweaks

### LINKED ISSUE #143: Simplify the use of `base` in the lobster configuration
We should simplify the use of the `base` path in storage configurations. Consider this configuration:
```
storage:
    base: /store/user/einstein/foo/bar
    local: /hadoop/store/user/einstein/foo/bar
    output: "chirp://earth.crc.nd.edu:9099/foo/bar"
```

This does not work because of the call to StorageElement.test() [here](https://github.com/matz-e/lobster/blob/master/lobster/se.py#L327). One way to fix this is to change the path passed to StorageElement.test() for Chirp() objects to remove the server part, is that the route we want to go? Or should the user use, e.g: `base: "chirp://earth.crc.nd.edu:9099//"`?

In principle, if only the `local` and `output` paths are given, they should be sufficient to completely specify the configuration. If possible we should eliminate the need for the user to specify `base` at all in such cases, unless it is providing unique information (for example, the user provides only `base` and `site`.) If we can't do that, we should clarify the documentation regarding the use of `base` (currently we only describe it in the context of configurations using the `site` identifier.)

## EDIT
We are almost finished with necessary modifications on the `se-tweaks` branch. As I see it (please add anything missing) there are two remaining items to finish this up:
- [x] finish testing 
    - [x] MC [Anna + Matthias]
    - [x] slimming, input from DBS [Matthias]
    - [x] slimming, local input [Anna]
- [x] update documentation
    - [x] README [Matthias]
    - [x] example configurations (ND examples) [Anna]
    - [x] example configurations (CERN examples) [Matthias]

matz-e:
Adjust storage element specification.  Fixes #143.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,13 +87,24 @@ CMS):
 
 3. Download an example configuration file and adjust it for your needs, e.g.:
 
+        # reads dataset from CMS DBS and produces slimmed copy
         wget --no-check-certificate \
-            https://raw.githubusercontent.com/matz-e/lobster/master/examples/slim.yaml \
+            https://raw.githubusercontent.com/matz-e/lobster/master/examples/slim_dbs.yaml \
             https://raw.githubusercontent.com/matz-e/lobster/master/examples/slim.py
+
+        # reads input files from specified directory and produces slimmed copy
+        wget --no-check-certificate \
+            https://raw.githubusercontent.com/matz-e/lobster/master/examples/slim_file_interface.yaml \
+            https://raw.githubusercontent.com/matz-e/lobster/master/examples/slim.py
+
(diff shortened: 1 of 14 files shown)


annawoodard:
Let's wait on support for local inputs before merging.

annawoodard:
@matz-e: we've marked off everything on our to-do list. There must be some odds and ends that we're missing, but I'd say `se-tweaks` is relatively stable and I'm ready to merge if you are.

matz-e:
Alright.  I had `nosetests` passing most SE tests except for the SRM protocol and a Chirp bug.

PR #172, opened 2015-08-07
Only start a process for auto logging if it is enabled. Fixes #171.

### LINKED ISSUE #171: Only start extra Lobster processes if they are needed
The extra process started for automatic plotting is kind of annoying until we fix #83. We shouldn't start it if it isn't going to be doing anything.

annawoodard:


matz-e:
Marvelous!

PR #174, opened 2015-08-07
Add new class for storing job update info. Fixes #173.

### LINKED ISSUE #173: Improve the readability of the job update container
We currently pass around a very long list of job-running stats to be updated in the job table. It is very cumbersome to add to and debug, as I was reminded of recently when adding access method stats to plotting. We should switch to something more readable in order to save time.

annawoodard:
Actually, it's a function which returns a generic class, so it could be used to make other containers for updating the database.

This should improve the readability and make it easier to add
new tracked quantities to the job table.

I made it more namedtuple-inspired than dictionary-inspired because all of our queries are currently constructed to take tuples rather than dictionaries. This approach would probably be more flexible if we switched to dictionary-like syntax + a dictionary-like container, but lets make that change later if it becomes an issue. For now, I just couldn't stomach having to pick through that long update list trying to figure out which index goes to which quantity ever again.

I'm not sure if util.py is the right place for it, I'll wait and see what @matz-e thinks.

### DIFF
--- a/lobster/cmssw/job.py
+++ b/lobster/cmssw/job.py
@@ -98,8 +98,7 @@ class JobHandler(object):
         file_update = []
         jobit_update = []
 
-        jobits_processed = 0
-        jobits_missed = 0
+        jobits_processed = len(self._jobits)
 
         for (id, file) in self._files:
             file_jobits = [tpl for tpl in self._jobits if tpl[1] == id]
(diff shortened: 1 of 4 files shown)


matz-e:
Looking good!  It will break the unit-tests, though, I think (did you run `nosetests`?)

annawoodard:
@matz-e: I've fixed the unit tests now, so this should be ready for review. (Sorry for the rebase and commit amend, which obscures the change. Next time I'll merge instead of rebase when the branch is already pushed.)

matz-e:
I like it!  I agree that we should move more things around into custom containers!

PR #177, opened 2015-08-27
Merge fail

klannon:
I'm new at this (first PR ever!) so tell me if I'm doing this wrong.  The changes I'd like to merge fix Lobster so that it will correctly handle merging of non-EDM files (using hadd instead of cmsRun).  I also made a few changes along the way to improve logging and make the code more resilient against failures.  (Basically, to determine the problem was non-EDM files in merging, I had to juggle some things so that the job.log made sense--issues with flushing stdout buffers meant some output was going in the wrong place--and also catch and handle an exception that was causing job.py on the worker to crash before zipping executable.log, which meant it was never transferred off the worker and I couldn't see what the real error message was.)  I guess now someone needs to look this over and say it's OK to merge?  I'll confess that although I did read the "guidelines for contributing" I don't know hose to use `nose` so I haven't tried any of the unit tests.  If that's something I should do, would someone give me a suggestion?

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,11 +117,12 @@ CMS):
    jobs finish.  If all processing is already done, only merge jobs will
    run.  Valid units for this option are `K`, `k`, `M`, `m`, `G`, and `g`.
 
-   To merge standard ROOT files, use
+   By default, Lobster will assume that the outputs of a task are in EDM
+   format.  To merge non-EDM ROOT files, specify
 
-        non edm output: true
+        edm output: false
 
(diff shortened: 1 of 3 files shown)


matz-e:
I'll have a closer look once I get back into the office.  How does this handle files that are not present, but should be merged?  And using `nose` just involves running `nosetests` in the directory you have your lobster source code in…

matz-e:
Thanks you, Kevin!

I've changed the option, I think it's a bit more logical that way.

PR #185, opened 2015-09-04
Hacked wrapper.sh to point X509_CERT_DIR and X509_VOMS_DIR at the gri…

klannon:
…d.cern.ch repository instead of the (currently) broken OSG one.  Hopefully this is temporary?

### DIFF
--- a/lobster/cmssw/data/wrapper.sh
+++ b/lobster/cmssw/data/wrapper.sh
@@ -120,6 +120,11 @@ if [[ -z "$LOBSTER_PROXY_INFO" || ( -z "$LOBSTER_LCG_CP" && -z "$LOBSTER_GFAL_CO
 	slc=$(egrep "Red Hat Enterprise|Scientific|CentOS" /etc/redhat-release | sed 's/.*[rR]elease \([0-9]*\).*/\1/')
 	source /cvmfs/oasis.opensciencegrid.org/osg-software/osg-wn-client/3.2/current/el$slc-$(uname -m)/setup.sh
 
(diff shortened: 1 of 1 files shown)


klannon:
I'm actually not sure whether we want to merge this or not, but I thought I'd exercise the pull request machinery.  This is, I think, at least an acceptable short-term solution to the x509 certs being broken in the OSG repository by pointing at the apparently updated CERN repository.  This could get fixed tomorrow, in which case, I'd argue for removing this hack.  Or it could linger for weeks.  What do you think?  The main motivation I see for merging this is so that people who follow the directions get a working version of lobster and not a broken one.

khurtado:
I think just merging and removing the hack once this issue is solved at the OSG side should be easy enough to do. I guess the only issue is if we all forget to remove the hack afterwards, of course :). On the other side, OSG people have 1 business day to reply to tickets and this issue should be very straightforward for them to fix, so I would probably wait at least today before merging.

I am pasting the OSG ticket related to this issue, so that we can check and know when it's solved.
https://ticket.opensciencegrid.org/26674

klannon:
Well, I just checked and the OSG is playing "pass the blame" with this right now.  The ticket has now been referred to three different people and no one has offered to help.  I'm thinking we should merge this.  I'm worried that the master branch is currently broken because of this, and I'm not confident it will be fixed soon.  Anyone object to a merge for this?

matz-e:
OK.  I think it would break running on some OSG sites (and CERN), but I don't forsee doing that soon…

PR #192, opened 2015-09-11
Fix srm test

annawoodard:


matz-e:
I would rather have it commented out with a reason, since otherwise you open up potential questions (why don't we test this?).

annawoodard:
@matz-e: Good point. I commented it out instead of removing it.

PR #194, opened 2015-09-16
Remove crab dependency

matz-e:
Putting this out for preliminary preview.  This should remove all dependencies on CRAB2.  Done by rewriting essential parts of the framework job report parsing (which is now only done for `cmsRun` jobs), and saving all information to our report JSON.  The latter is the only file needed for publishing, merging, etc...

Testing is still ongoing.


matz-e:
First run with my ttH analysis looked OK.  I'm still doing some more testing.

annawoodard:
Awesome!

PR #195, opened 2015-09-17
Better task handling

### LINKED ISSUE #155: Improve Handling of Problem Files
Right now, if there are a few bad input files (e.g. corrupted at a T2 site), Lobster tries forever on those files.  It would be nice to tell Lobster to give up after a certain number of failures and move on to finish the rest of the jobs.  Here's what would be needed basically to implement the minimal version:

From @matz-e on [Slack](https://ndlobster.slack.com/messages/lobster-debug/):
>to account for skipped jobits, jobits_done in jobit.py would have take into account the jobits skipped from the files table (where appropriate.) line 213 is where the files would have to be actually skipped during job creation.  Then we would probably like to include skipped jobits in the plotting info page.

There are a number of additional features which should be included to make a complete implementation:
From @annawoodard on [Slack](https://ndlobster.slack.com/messages/lobster-debug/):
>And you'll need to add a flag in the lobster configuration that the user can use to set the number of retries before skipping a file 

and

>You should also probably add a snippet in cmssw/data/job.py which ensures that skipBadFiles is enabled 

From me:
* I think we also need some way for the user to specify whether skipping files is OK.  (E.g. for data we might not want to skip files.  
* Ideally, there'd also be a way to come back later and say please retry those skipped files again without expecting the user to hack the DB directly.

### LINKED ISSUE #186: Skip Problematic Jobs or Jobits and Move on to Merging
This is a critical feature for day-to-day usability:  If there is some piece of a dataset that cannot be processed (i.e. a corrupted file which causes your job to crash).  There needs to be a way either to tell Lobster to skip that piece, or for Lobster to figure out automatically that it's crashed the same way many times on a particular piece of work and shouldn't keep trying any more.

I think in this case, since users don't generally interact with specific jobs in Lobster, Lobster should decide when it's going to skip something, rather than the user trying to blacklist individual jobits manually.  (Although it might be handy to be able to blacklist specific input files...)  So, I think it would be nice if we could have a retry threshold, where if a given jobit has failed too many times, Lobster marks it as skipped and moves on.

In my ideal implementation, it would be that when a jobit reaches a given number of failures (e.g. non-zero return codes, eviction of course shouldn't count), Lobster tries the jobit one last time, but in a job all by itself.  This allows Lobster to process jobits who were only failing because they were consistently being grouped with one bad jobit.  (This certainly happens when there are so few jobits left that the same ones keep getting put together into one job.)  If the jobit fails after that, then it gets marked as skipped, and lobster moves on--merging all non-skipped job output and concluding its task.

There needs also to be a way for users to clear the skipped flags and let lobster try to finish the jobs that it previously skipped.

Is this feasible?  Suggestions on how to achieve it?  Better ideas?

matz-e:
Sits on top of #194, and adds simple task tapering when the number of requested tasks exceeds the number of tasks left to create.  Also tries to scale the task size to a user configurable maximum runtime, telling CMSSW to quit after an additional grace period of 10 minutes.

Also keeps failed jobits and skipped files in check and thus fixes #186, and fixes #155.

### DIFF
--- a/doc/CERN.md
+++ b/doc/CERN.md
@@ -18,8 +18,6 @@ version of the CC tools!
 Adjust your paths as follows:
 
     export PYTHONPATH=/afs/cern.ch/user/m/matze/src/cctools-002-b7289441-cvmfs-90fca639c-x86_64-redhat6/lib/python2.6/site-packages/:$PYTHONPATH
-    export PYTHONPATH=/cvmfs/cms.cern.ch/crab/CRAB_2_11_1/external:$PYTHONPATH
-    export PYTHONPATH=/cvmfs/cms.cern.ch/crab/CRAB_2_11_1/python:$PYTHONPATH
     export PATH=$PATH:/afs/cern.ch/user/m/matze/src/cctools-002-b7289441-cvmfs-90fca639c-x86_64-redhat6/bin
 
 ## Actually running
--- a/examples/beanprod.yaml
+++ b/examples/beanprod.yaml
@@ -28,8 +28,20 @@ delete merged: true
 ## optional: if a job exits with a code in this list, the host it ran on will be blacklisted
 # bad exit codes: [1, 2]
 
+## path to CMSSW release needed to run executable (default read from $LOCALRT)
+# sandbox release top: /path/to/release
+
+## specify sandbox instead of creating one at startup
+# recycle sandbox: /path/to/sandbox.tar.bz2
+
 ## optional: specify directories in CMSSW work area to omit from sandbox
 # sandbox blacklist: ['*DrawPlots*']
+
+## DBS instance to publish to
+# dbs instance: phys03
+
+## overwrite the username to use for publishing, in case it is different from the environment var $USER
+# publish user: username
 
 advanced:
   ## start killing jobs with excessive runtimes after this many successful
@@ -45,13 +57,6 @@ advanced:
   log level: 0
   ## how many jobs to create and keep in the queue
   # payload: 400
-
-## DBS instance to publish to
-# dbs instance: phys03
-## overwrite the username to use for publishing, in case it is different from the environment var $USER
-# publish user: username
-## specify sandbox instead of creating one at startup
-# recycle sandbox: path/to/sandbox.tar.bz2
 
 task defaults:
     cmssw config: makeBEAN_cfg.py
--- a/examples/mc_production.yaml
+++ b/examples/mc_production.yaml
@@ -35,8 +35,20 @@ delete merged: true
 ## optional: if a job exits with a code in this list, the host it ran on will be blacklisted
 # bad exit codes: [1, 2]
 
+## path to CMSSW release needed to run executable (default read from $LOCALRT)
+# sandbox release top: /path/to/release
+
+## specify sandbox instead of creating one at startup
+# recycle sandbox: /path/to/sandbox.tar.bz2
+
 ## optional: specify directories in CMSSW work area to omit from sandbox
 # sandbox blacklist: ['*DrawPlots*']
+
+## DBS instance to publish to
+# dbs instance: phys03
+
+## overwrite the username to use for publishing, in case it is different from the environment var $USER
+# publish user: username
 
 advanced:
   ## start killing jobs with excessive runtimes after this many successful
@@ -50,13 +62,6 @@ advanced:
   log level: 0
   ## how many jobs to create and keep in the queue
   # payload: 400
-
-## DBS instance to publish to
-# dbs instance: phys03
-## overwrite the username to use for publishing, in case it is different from the environment var $USER
-# publish user: username
-## specify sandbox instead of creating one at startup
-# recycle sandbox: path/to/sandbox.tar.bz2
 
 task defaults:
     cmssw config: mc_production.py
--- a/examples/slim_dbs.yaml
+++ b/examples/slim_dbs.yaml
@@ -32,8 +32,20 @@ delete merged: true
 ## optional: if a job exits with a code in this list, the host it ran on will be blacklisted
 # bad exit codes: [1, 2]
 
+## path to CMSSW release needed to run executable (default read from $LOCALRT)
+# sandbox release top: /path/to/release
+
+## specify sandbox instead of creating one at startup
+# recycle sandbox: /path/to/sandbox.tar.bz2
+
 ## optional: specify directories in CMSSW work area to omit from sandbox
 # sandbox blacklist: ['*DrawPlots*']
+
+## DBS instance to publish to
+# dbs instance: phys03
+
+## overwrite the username to use for publishing, in case it is different from the environment var $USER
+# publish user: username
 
 # advanced:
   ## start killing jobs with excessive runtimes after this many successful
@@ -49,13 +61,6 @@ delete merged: true
   # log level: 0
   ## how many jobs to create and keep in the queue
   # payload: 400
-
-## DBS instance to publish to
-# dbs instance: phys03
-## overwrite the username to use for publishing, in case it is different from the environment var $USER
-# publish user: username
-## specify sandbox instead of creating one at startup
-# recycle sandbox: path/to/sandbox.tar.bz2
 
 ## task fields:
 ##      required:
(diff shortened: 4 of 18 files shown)


klannon:
How's this for testing--something only a guy like me could provide!  If you use this branch and start a project, but don't include `task runtime` in your config, then all your jobs crash (exit code 1) with the following error:
```
  File "job.py", line 637, in <module>
    edit_process_source(pset_mod, config)
  File "job.py", line 426, in edit_process_source
    runtime = config['task runtime']
KeyError: 'task runtime'
```
I assume this is a bug, but if I'm wrong and it's a feature (desired even?) then let me know.  Is it possible to assign a default `task runtime` of "unlimited" to jobs that don't specify?  If it's essential that all tasks define a runtime, is it possible for Lobster to detect when this is missing and abort the run, rather than generating doomed tasks at the maximum possible rate?

annawoodard:
I think we definitely want to assign a default, especially considering `maxSecondsUntilRampdown` is only supported in CMSSW > 7.4.

matz-e:
I ran this config with `CMSSW_5_3_18`:
```
import FWCore.ParameterSet.Config as cms

process = cms.Process('slim')

process.source = cms.Source("PoolSource", fileNames = cms.untracked.vstring())
process.maxEvents = cms.untracked.PSet(input = cms.untracked.int32(10))
process.output = cms.OutputModule("PoolOutputModule",
    outputCommands = cms.untracked.vstring("drop *", "keep recoTracks_*_*_*"),
    fileName = cms.untracked.string('output.root'),
)

process.maxSecondsUntilRampdown = cms.untracked.PSet(input = cms.untracked.int32(12))

process.source = cms.Source("PoolSource",
        fileNames = cms.untracked.vstring([
            '/store/mc/RunIISpring15DR74/ttHJetToTT_M125_13TeV_amcatnloFXFX_madspin_pythia8/MINIAODSIM/Asympt25ns_MCRUN2_74_V9_ext1-v1/00000/126F72B6-E407-E511-AC78-002481E75CF0.root'
        ])
)


process.out = cms.EndPath(process.output)
```
`cmsRun` parses the config and fails opening the file.  Either that's transient XrootD, or XrootD has changed since the release?  In any case, I think the parameter is just ignored, so no harm done.

annawoodard:
Hmm, I'm not convinced by tests that fail for other reasons, because it's not always clear what order things will fail in. :)

In any case, I tested with the mc production example config, and it looks fine to me too.

matz-e:
Well, I've had time to run the following config:
```
import FWCore.ParameterSet.Config as cms

process = cms.Process('slim')

process.source = cms.Source("PoolSource", fileNames = cms.untracked.vstring())
process.maxEvents = cms.untracked.PSet(input = cms.untracked.int32(10))
process.output = cms.OutputModule("PoolOutputModule",
    outputCommands = cms.untracked.vstring("drop *", "keep TriggerResults_*_*_*"),
    fileName = cms.untracked.string('output.root'),
)

process.maxSecondsUntilRampdown = cms.untracked.PSet(input = cms.untracked.int32(12))

process.source = cms.Source("PoolSource",
        fileNames = cms.untracked.vstring([
            'file:/hadoop/store/mc/Summer11/DYToMuMu_M-20_TuneZ2_7TeV-pythia6/AODSIM/PU_S3_START42_V11-v2/0000/B4C7856F-F37C-E011-97A9-002481E94C56.root'
        ])
)


process.out = cms.EndPath(process.output)
```
and it runs just fine with `CMSSW_5_3_18`.

annawoodard:
Using `CMSSW_5_3_18` (ships with python 2.6), I'm having trouble with the new dependencies. I'm not sure if it's more straightforward to try and sort out the dependencies, or just say that we don't support python 2.6 anymore. We could require the user to run lobster with a release after the switch to 2.7, and allow them to override `CMSSW_BASE` in their lobster config in order to pack up the sandbox for releases using 2.6. Thoughts, @matz-e or @klannon?

```
Using /afs/crc.nd.edu/user/a/awoodard/.local/lib/python2.6/site-packages/MarkupSafe-0.23-py2.6.egg
Finished processing dependencies for Lobster==1.3
Error in atexit._run_exitfuncs:
  File "/cvmfs/cms.cern.ch/slc6_amd64_gcc472/external/python/2.6.4/lib/python2.6/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/cvmfs/cms.cern.ch/slc6_amd64_gcc472/external/python/2.6.4/lib/python2.6/multiprocessing/util.py", line 258, in _exit_function
    info('process shutting down')
TypeError: 'NoneType' object is not callable
Error in sys.exitfunc:
  File "/cvmfs/cms.cern.ch/slc6_amd64_gcc472/external/python/2.6.4/lib/python2.6/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/cvmfs/cms.cern.ch/slc6_amd64_gcc472/external/python/2.6.4/lib/python2.6/multiprocessing/util.py", line 258, in _exit_function
    info('process shutting down')
TypeError: 'NoneType' object is not callable
```

klannon:
Well, I think you are our main 2.6 user, so whatever makes your life easiest (e.g. allows you to run what you need with the least amount of fuss.)

annawoodard:
I think we'll always have issues where one group of users is adopting newer versions faster than other groups. So, while not as elegant, I think it makes sense to de-couple the `CMSSW` version used to run lobster with the `CMSSW` version used for the user's analysis code. In CRAB3 it's been a constant struggle to keep the CRAB3/CMSSW environments playing nicely, but I don't see it being very onerous to just have two terminal windows open. Using this approach, we can even recommend a 'tested' CMSSW version in our installation instructions, which seems more stable than catching this type of issue as it comes up.

annawoodard:
I can add in a `CMSSW_BASE` override, anyone want to opine on what the clearest field name is for the lobster configuration?

matz-e:
I like that.  grid-control had that, too.  `cmssw base` or so in the config?

matz-e:
Also, we could force the version of python that lobster is executed with.  Have you tried installing lobster with 74 or so and then switching releases?

matz-e:
This is what is set as python executable in `.local/bin/lobster`: `/cvmfs/cms.cern.ch/slc6_amd64_gcc491/cms/cmssw/CMSSW_7_4_7/external/slc6_amd64_gcc491/bin/python`

So if you install in 74 or similar, you should always execute the installed lobster in python 2.7.

matz-e:
Of course, for running a local development version, you could still use a `CMSSW_BASE` override.

klannon:
Is the problem just the python version, or are there other issues with CMSSW version?  Also, are you suggesting that we tie Lobster to a specifc CMSSW version?  It seems like you can probably count on it working OK for a range (e.g. 7X or 74X)?  Or am I not understanding you...

annawoodard:
Yes, using anything shipping with 2.7 works for me. I just mentioned the issue because we need the override to do work with older analysis code.

> Of course, for running a local development version, you could still use a CMSSW_BASE override.

I don't understand this. As far as I know, we only need the override to pack up the sandbox correctly. I should be able to just switch to 2.7 for all my development versions.

annawoodard:
I'm not suggesting we tie it to a specific version, I'm just saying we can simplify things by not supporting 2.6 anymore.

matz-e:
> I don't understand this. As far as I know, we only need the override to pack up the sandbox correctly. I should be able to just switch to 2.7 for all my development versions.

That's what I mean, sorry…

klannon:
OK, I think we're in complete agreement.  Lobster is 2.7 only.  We tell people which version of CMSSW has been used to test it.  Other versions can be used at own risk.  We support submitting projects that use a different CMSSW version via CMSSW_BASE override.  That all sounds great to me.

matz-e:
For lobster installed via `pip` and located in `.local/bin`, executing lobster after `cmsenv` in 5x releases should just work, no need to override things.  Running `my-git-checkout/lobster.py` needs it - we should make that distinction in the documentation.

annawoodard:
Hmm. It could break at anytime though, so I still vote that we drop support for 2.6 and instead of having two sets of instructions (one for pip installation and one for `my-git-checkout/lobster.py`, just one set (use 2.7 for running lobster, if your release uses <2.7, then set `cmssw base` in your lobster configuration.

matz-e:
OK!

klannon:
+1 for dropping 2.6 support but also +1 for explaining any differences between pip and git installs in documentation!

annawoodard:
I've also tested this with a file-based, non-cmssw configuration. Looks great!

PR #198, opened 2015-09-21
Allow command arguments

annawoodard:


matz-e:
Awesome!

matz-e:
(good one with the `basestring`, too!)

PR #204, opened 2015-09-24
Fix backend tests

annawoodard:


matz-e:
Thank you so much for this!  You rock!

Issue #28, opened 2014-07-02
Syncronize configuration with stored one


matz-e:
We should offer the possibility to update settings based on an updated config passed to lobster.  This includes cross-checking which settings can be altered without any double-checking, which need confirmation by the user, and which should not be changed.

klannon:
Are you thinking something like a `lobster update config.yaml` command, which would give Lobster a chance to validate the new config along the lines of #13 and what you stated above.  If the config is good and the user hasn't tried to change any illegal parameters, then `lobster update` can just replace the old config in the working directory with the updated one.  Otherwise, it can give the user a helpful (or at least explicit) message saying the config has a problem and it won't be updated.

Presumably, Lobster doesn't currently re-read its config periodically.  If we implement this, we'd probably also want to have Lobster occasionally re-read its config, right?

matz-e:
I'm actually thinking about something more complex. I have a branch open which turns the whole config into a python object, which allows for config verification (would be a real pain with YAML), and has a more fine-grained split of settings.

Then I think instead of re-reading the config, we should update settings directly via IPC. Comparing two configs is another issue that seems quite complicated to get right, and then propagating the changes to the right python variable is needed. If everything is in one python object tree, modifications will be available immediately.

klannon:
OK, well it seems that you're deeper into this than I can hope to be in the near future.  My only request is that changing a running lobster config should require some explicit action by the user (e.g. it shouldn't be an implicit side-effect of editing a config) and if the change won't be propagated to the master, the user should get unambiguous feedback to that effect.  Nothing is more frustrating when debugging than thinking you've made some change that never actually propagated because of some silly mistake.

matz-e:
Yes, it'll be explicit. And I understand that there should be feedback about any value change. They should actually be logged and reported back to the user on the command line.

matz-e:
I've implemented it such that feedback is provided in the main Lobster log. Two-way communication via unix utilities without the need of a full blown internet socket is not quite feasible.  Sample change command:
```
(.lobster)earth: ./lobster.py reconfigure workdir 23456 examples/test2.yaml
2016-01-20 07:05:11 [INFO] lobster.reconfigure: sending command: config.workdir = 23456
2016-01-20 07:05:11 [INFO] lobster.reconfigure: check the log of the main process for success
```
And Lobster processing log:
```
2016-01-20 07:05:28 [DEBUG] lobster.actions: received commands: config.workdir = 23456
[...]

PR #205, opened 2015-09-25
User fixes

### LINKED ISSUE #77: Need to be able to resume moved working directories
Currently, lobster fails in all kinds of ways when a user moves the working directory and tries to resume working with it, e.g., by plotting from it.

### LINKED ISSUE #84: Make a copy of extra inputs.
We currently keep a copy of the cmssw pset, but not any other extra inputs which are specified in the configuration. I think that we should do so. If there is a good reason not to, then we should at least save the full path to those files. This is related to #77, but I'd rather keep it separate.

matz-e:
Just putting this out there.  Validation is broken if the user does not specify output files in the configuration.  This will save an expanded configuration to the working directory, containing output files and global tags.

For complete user-friendliness, the following still have to be addressed, at least partially:

- [ ] #28 
- [x] Fixes #84 
- [x] Fixes #77

### DIFF
--- a/examples/custom_command.sh
+++ b/examples/custom_command.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+echo arguments: $*
+
+optional=
+output=
+
+while getopts a:o: opt; do
+	case $opt in
+		a)
+			optional=$OPTARG
+			;;
+		o)
+			output=$OPTARG
+			;;
+		?)
+			exit 1
+			;;
+	esac
+done
+
+for i in $(seq $(($OPTIND-1))); do
+	shift
+done
+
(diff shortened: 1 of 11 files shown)


annawoodard:
@matz-e: do you want this reviewed and merged now, or after the to-do list is finished?

matz-e:
I'd prefer to resolve at least #77, still. #28 is a hard one!

matz-e:
@annawoodard: I think this should be ready, #28 is too big in scope to be really addressed here? Note that changes to the configuration need to be done to the configuration in the working directory, not the one passed on the command line.

annawoodard:
Looks great!

PR #211, opened 2015-10-02
Handler shuffle

### LINKED ISSUE #67: Move more code to the jobhandler
A lot of the code involved with creating and returning jobs could be moved to the handlers.  There, it should be grouped by purpose, maybe called via call-table?  Sort of like the pset handling in crab is organized, that we basically have a configuration and, depending on job type (process, merge, cms, file-based) call modifying methods on that.

matz-e:
Meant to eventually fix #67. Our current structure with tons of `if` statements when dealing with tasks is not really sustainable, and there's very little overview of which options can actually be passed for a workflow.

This PR should simplify how we handle task creation (and returning) by

* introducing a `Workflow` class for workflow-wide settings (which are still referred to as `task` in our configuration)
* moving the `JobHandler` to separate file

The `obtain` method of the `JobSource` should read as follows:

1. Get a bunch of stuff from the database
2. Create a skeleton of parameters we want to send along with the task
3. Have the `Workflow` apply workflow-wide settings
4. Have the `TaskHandler` (formerly known as `JobHandler`) apply task-specific settings
5. Send it off!

There are still some more code-rearrangements to come…

### DIFF
--- a/lobster/__init__.py
+++ b/lobster/__init__.py
@@ -1,1 +1,1 @@
-
+import cmssw
--- a/lobster/cmssw/__init__.py
+++ b/lobster/cmssw/__init__.py
@@ -1,5 +1,4 @@
 from actions import Actions
+from handler import TaskHandler
+from workflow import Workflow
 from job import JobProvider
-from jobit import JobitStore
-from plotting import plot
-from publish import publish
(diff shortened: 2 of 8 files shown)


annawoodard:
Awesome!

annawoodard:
I'm going to merge this now, future code re-arrangements can be part of another PR.

PR #218, opened 2015-10-15
Provide the user with a finalize option to wrap up processing.

matz-e:
@annawoodard, @klannon I think we should provide this as a user option.


annawoodard:
Looks good to me.

PR #240, opened 2015-11-19
More code shuffle

matz-e:
I think we should merge this, and then continue branching off on top of it again. Otherwise, I think that the development is straying a little far from the master branch. The only big change I could still think of would be renaming `cmssw` to `cms` in the directory structure, but I'm personally on the fence about this.


annawoodard:
Agreed!

PR #248, opened 2015-11-23
Plot fixes

matz-e:
A bunch of plotting stuff. Still WIP and depends on #240.


annawoodard:
Looks great! HTML, here we come!

PR #262, opened 2016-01-05
Plot by workflow

matz-e:
Adds a dropdown to the monitoring to display task plots for a single workflow only. Useful to separate out failure modes, memory consumption and task runtimes.

Still could need some fine-tuning, as the selected workflow currently resets when the page reloads (every 15 minutes).

* [x] split plotting page by workflow (produce several pages), providing a Lobster-only overview. Also reduces JS again to the minimum needed. Either we should have a fully dynamic page (with a server backend), or provide more simple pages.


matz-e:
I think this is fairly finished. Now uses separate pages for the Lobster overview and (one or all) workflows. Further fine-tuning would be good, but could be done in the master.

annawoodard:
Looks great!

PR #263, opened 2016-01-06
Allow to specify memory requirements per task.

### LINKED ISSUE #255: Allow to specify per workflow resource requirements
Current Lobster usage involves one contract, normally: we tell the pool what resources to request and place no other requirements on tasks but the number of cores they use. With more heterogenous projects involving different workflows, it would be better to specify the resources per workflow, and optionally let WQ lower/adjust the resource estimates as needed.

matz-e:
Use the setting `task memory` on a per workflow basis to do so.
Fixes #255, may need minor adjustment for merge tasks.

### DIFF
--- a/lobster/commands/process.py
+++ b/lobster/commands/process.py
@@ -245,16 +245,15 @@ def sprint(config, workdir, cmstask):
                 break
 
             hunger -= len(tasks)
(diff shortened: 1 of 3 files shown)


matz-e:
I think this could be merged, @annawoodard. The new WQ stuff builds on top of it.

PR #268, opened 2016-01-07
Update to latest WQ

matz-e:
This updates us to the latest and greatest WQ, which tailors the resource allotment itself. We provide initial estimates via the user, and WQ will adjust these to the reality.


matz-e:
I think this should be ready to get merged?

annawoodard:
Looks good!

PR #282, opened 2016-02-01
Python config (and much more)

### LINKED ISSUE #13: Check for valid configuration details
We need to check
* that a task contains all required information
* that all labels are unique and match `[A-Za-z][A-Za-z0-9_]*`

### LINKED ISSUE #28: Syncronize configuration with stored one

### LINKED ISSUE #79: Allow python based task configuration
Switching to python to generate tasks would allow us to have a more dynamic configuration of tasks, and render any scripts to generate configurations moot.  This would probably be pretty handy to have for final analysis steps (at least in my case).  Example:
```
from lobster.cmssw import Task

proto = Task(…)
toplevel = …

for sys in "cms_ttH_jes cms_ttH_qSquared".split():
    for dir in ("Up", "Down"):
        t = proto.clone()
        t.args += ['--sys', sys + dir, '--outfile', os.path.join(toplevel, sys  + dir, 'foo.root')]
```
where creating a `Task` object automatically registers it in a class-wide registry.

### LINKED ISSUE #271: Multi-core tasks put too many tasks in the queue
With the current task generation, Lobster tries to fill up all uncommitted cores and have a buffer of tasks to fill up 10% of the total cores. With multi-core tasks, we overproduce tasks. Should be an easy fix, just some additional arithmetic.

### LINKED ISSUE #275: Need Way to Cap Number of Running Jobs For Specific Workflows
Digi completely saturates the campus bandwidth, but Lobster chooses to run only that because of the distribution of jobs.  Need a way to tell it don't run more than XX jobs of this type so it will continue to make progress on other steps as it can.  (There's still plenty of GS to do, but it choose digi because that's the step with the most available work to do, I think...)

### LINKED ISSUE #276: Need to refresh Lobster settings while running
Should be done by opening a socket in the Lobster working dir, and reading values from that periodically to adjust, with a simple protocol. A Lobster command could then talk to the socket, interfacing with the user. After changing values, the resulting changes should be re-pickled.

matz-e:
Sorry for the ginormous code-dump…

I've patched this together starting before christmas… a short overview over the changes:

* the configuration is mainly in python now
  * allows to verify the configuration, since most settings are passed to the constructor of python objects
  * YAML can be converted to python (not sure how long that will work)
  * documentation is where the code is – should make it harder to document settings
* inter-process communication and changing settings via `lobster reconfigure setting value config`
* auto-generated documentation to a website via sphinx (still WIP), breaking things into more manageable chunks than with one gigantic README

Fixes #79, fixes #13, fixes #28 (for selected values), fixes #271, fixes #275, fixes #276, and will allow to fix a whole bunch more.


annawoodard:
I'm not going to be able to do a thorough code review of this within the next few days, so I'm going ahead and merging. Skimming things this all looks great.

PR #287, opened 2016-02-18
Scale 2015 01 v2

klannon:
Feel free to delete this PR and proceed in a different way.  What I was going for here is to make sure the last two commits to the scale test branch get merged in, assuming you're comfortable with what I did there.  See 29a3f40 and 401ff62.

### DIFF
--- a/doc/ND.md
+++ b/doc/ND.md
@@ -7,13 +7,13 @@ your setup might be needed.
 
 Use `work_queue` etc from the CC lab:
 
-    cctools=cctools-047-4b636166-cvmfs-40cf5bba
+    cctools=053-15727c22-cvmfs-40cf5bba
     export PYTHONPATH=$PYTHONPATH:/afs/crc.nd.edu/group/ccl/software/$cctools/x86_64/redhat6/lib/python2.6/site-packages
     export PATH=/afs/crc.nd.edu/group/ccl/software/$cctools/x86_64/redhat6/bin:$PATH
 
 or, for `tcsh` users,
 
-    setenv cctools-047-4b636166-cvmfs-40cf5bba
+    setenv 053-15727c22-cvmfs-40cf5bba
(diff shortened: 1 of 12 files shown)


matz-e:
I think I've actually cherry-picked most of your commits on top of the master (since this branch has had merge-conflicts for a while)… I think that 29a3f40 is the only one that's not in the master yet, and I'm still weighing the changes – if we take out sizes for file-based datasets, do we need them at all? I think the fast grabbing of `.root` files may be reasonable. Of course these changes will not apply cleanly anymore, so I'll have to migrate them to the appropriate places.

annawoodard:
> if we take out sizes for file-based datasets, do we need them at all?

@matz-e: I can't think of a reason why we would. Maybe we should just open this as a separate issue, to apply 29a3f40 and take out file sizes for file-based datasets.

matz-e:
I guess it would give us a rough estimate of how much data needs to flow over the network, especially from outside via XrootD. But we could still remove it for local datasets, as that number can easily be gotten by a file-system query.

annawoodard:
> But we could still remove it for local datasets, as that number can easily be gotten by a file-system query.

@matz-e, @klannon: I vote for local and nonlocal datasets to be on equal footing. I also vote that `.root` and non-`.root` files also be on equal footing (which doesn't agree with 29a3f40.) What makes sense to me to be on different footing is whether or not we are using a file list-- that determines whether or not the filesize info is free (it is free if we aren't using a file list.) Since we've only seen a case where getting the info causes a problem and not where it solves a problem, I think we should consider taking it out when we use a file list. If we ever come to the point where we feel like we needed that info and didn't have it, we can open a new issue to come up with a more careful treatment.

klannon:
For posterity:  If we really wanted to do this, we can't query each file individually because of the overhead.  If we reworked things to do the operation in batches then it becomes manageable.  But it would still require care because I think in some instances there is a maximum number of files you can pass into a command (e.g. when you're just doing the operations via subprocess to a shell that has a max number of command line arguments limitation.)

PR #293, opened 2016-03-08
Add ca_info to DbsApi call 

annawoodard:
I was overcomplicating things before-- turns out we can just use the usual methods to obtain the proxy for the DbsApi.


matz-e:
Awesome!

PR #295, opened 2016-03-11
Fix setup and move docs.

### LINKED ISSUE #294: Document required numpy version
```
==> /tmpscratch/users/awoodard/lhe_23d_v18a/process.err <==
  File "/afs/crc.nd.edu/user/a/awoodard/lobster/lobster/actions.py", line 44, in plotf
    plotter.make_plots(foremen=config.foremen_logs)
  File "/afs/crc.nd.edu/user/a/awoodard/lobster/lobster/commands/plot.py", line 963, in make_plots
    bad_hosts=self.find_failure_hosts(failed_tasks),
  File "/afs/crc.nd.edu/user/a/awoodard/lobster/lobster/commands/plot.py", line 593, in find_failure_hosts
    hosts, counts = np.unique(failed_tasks['host'], return_counts=True)
TypeError: unique() got an unexpected keyword argument 'return_counts'
```
CMSSW_7_4_15 sets up numpy 1.6 for me, It looks like the 'return_counts' keyword was introduced in numpy 1.9.

I see that I can access numpy 1.9 at ND here `/afs/crc.nd.edu/x86_64_linux/python/2.7.8/lib/python2.7/site-packages/numpy` by loading the module but, with non-ND collaborators coming on board, I think we should try to document a more general way of accessing the required version. I agree with previous conversations that it is beyond our scope right now to completely decouple from CMSSW... Are there CMSSW releases that are setting up numpy 1.9? If so, we can just recommend them in the docs. Where are you pulling numpy from, @matze?

matz-e:
Fixes #294 by specifying the `numpy` version in `setup.py`. Also moves the documentation to a more canonical location. Readthedocs should find this automatically.


matz-e:
Should mention that I tested this with 74x, and installing numpy and matplotlib posed no problem with a fresh virtualenv.

annawoodard:
You got numpy 1.9 with 74x?

In any case, this looks great, thanks!

matz-e:
I was able to install and use numpy in 74x…

matz-e:
Hm. I need to go back and check something…

PR #296, opened 2016-03-15
Dump config

matz-e:
Changes the commands to a plugin-based infrastructure. Adding new commands requires only changes to a single file implementing the command in the `commands` directory. Do so to add a `configuration` command.

### DIFF
--- a/lobster/util.py
+++ b/lobster/util.py
@@ -34,7 +34,7 @@ class PartiallyMutable(type):
             args = inspect.getargspec(cls.__init__).args
             for attr in attrs['_mutable']:
                 if attr not in args:
-                    raise AttributeError('class {} defines {} as mutable, but does not list it in the constructor'.format(name))
+                    raise AttributeError('class {} defines {} as mutable, but does not list it in the constructor'.format(name, attr))
         elif len(attrs['_mutable']) > 0:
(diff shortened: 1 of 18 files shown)


lobster/util.py
```diff
@@ -29,6 +30,13 @@ def __init__(cls, name, bases, attrs):
             raise AttributeError('class {} does not set the attribute _mutable'.format(name))
         elif not isinstance(attrs[key], dict):
             raise AttributeError('class {} does not define the attribute _mutable as dict'.format(name))
+        if '__init__' in attrs:
+            args = inspect.getargspec(cls.__init__).args
+            for attr in attrs['_mutable']:
+                if attr not in args:
+                    raise AttributeError('class {} defines {} as mutable, but does not list it in the constructor'.format(name))
```
annawoodard: I think this is missing an argument, should be: `'class {} defines {} as mutable, but does not list it in the constructor'.format(name, attr)`

matz-e: Fixed, thanks!

annawoodard:
This is great. I love that all of the constructor arguments now must agree with property names. I think I'll enforce that the names agree when I'm saving arguments as properties in my coding from now on, even when not doing fancy metaprogramming; I think it makes the code easier to understand.

Also, :+1: for the inspect module, it was fun learning about that too!

PR #298, opened 2016-03-18
Remove hardcoded ndcms

### LINKED ISSUE #15: Remove Notre Dame specific hard-coded bits
The following are hard-coded and should be configurable:

* `parrot` binary
* the local site `T3_US_NotreDame`
* the http proxy

We should make these either configurable via the master or pass them along in the environment of `condor_submit_workers`.  I would prefer the latter for the last two poins, as this would allow to run jobs at different sites, using local caches.

### LINKED ISSUE #297: We should remove hardcoded instances of `ndcms.crc.nd.edu`
This has been changed: ndcms.crc.nd.edu -> deepthought.crc.nd.edu. We should remove hardcoded instances of the CE.

As @matz-e pointed out: 
[https://github.com/matz-e/lobster/search?utf8=%E2%9C%93&q=ndcms.crc.nd.edu](https://github.com/matz-e/lobster/search?utf8=%E2%9C%93&q=ndcms.crc.nd.edu)

matz-e:
This goes a bit towards #15 by removing mentions of NDCMS servers, and instead using WMCore stuff to obtain Frontier proxies, and attempting to get the CVMFS proxy from the canonical configuration file. Still fairly experimental (tests are running), but should fix #297, too.

Fixes #15, too, by now.

### DIFF
--- a/lobster/cmssw/dash.py
+++ b/lobster/cmssw/dash.py
@@ -89,7 +89,7 @@ class Monitor(DummyMonitor):
             self._ce = loadSiteLocalConfig().siteName
         except SiteConfigError:
             logger.error("can't load siteconfig, defaulting to hostname")
-            self._ce = socket.gethostname()
+            self._ce = socket.getfqdn()
 
     def __del__(self):
         self.free()
--- a/lobster/cmssw/dash.py
+++ b/lobster/cmssw/dash.py
@@ -47,7 +47,7 @@ class DummyMonitor(object):
         self._workflowid = util.checkpoint(workdir, 'id')
 
     def generate_ids(self, taskid):
-        return str(taskid), str(taskid)
+        return "dummy", "dummy"
 
     def register_run(self):
         pass
--- a/lobster/cmssw/dash.py
+++ b/lobster/cmssw/dash.py
@@ -165,7 +165,10 @@ class Monitor(DummyMonitor):
             'StatusValue': status,
             'StatusEnterTime':
             "{0:%F_%T}".format(datetime.datetime.utcnow()),
-            'StatusDestination': self._ce,
+            # Destination will be updated by the task once it sends a dashboard update.
+            # in line with
+            # https://github.com/dmwm/WMCore/blob/6f3570a741779d209f0f720647642d51b64845da/src/python/WMCore/Services/Dashboard/DashboardReporter.py#L136
+            'StatusDestination': 'Unknown',
             'RBname': 'condor'
             })
 
--- a/lobster/cmssw/dash.py
+++ b/lobster/cmssw/dash.py
@@ -1,12 +1,14 @@
 import datetime
 import logging
 import os
+import socket
 import subprocess
 
 from hashlib import sha1
 
 from WMCore.Services.Dashboard.DashboardAPI import apmonSend, apmonFree
 from WMCore.Services.SiteDB.SiteDB import SiteDBJSON
+from WMCore.Storage.SiteLocalConfig import loadSiteLocalConfig, SiteConfigError
 from lobster import util
 
 import time
@@ -45,10 +47,7 @@ class DummyMonitor(object):
         self._workflowid = util.checkpoint(workdir, 'id')
 
     def generate_ids(self, taskid):
-        monitorid = '{0}_{1}/{0}'.format(taskid, 'https://ndcms.crc.nd.edu/{0}'.format(sha1(self._workflowid).hexdigest()[-16:]))
-        syncid = 'https://ndcms.crc.nd.edu//{0}//12345.{1}'.format(self._workflowid, taskid)
-
-        return monitorid, syncid
+        return str(taskid), str(taskid)
 
     def register_run(self):
         pass
@@ -86,6 +85,12 @@ class Monitor(DummyMonitor):
         else:
             self.__executable = 'Unknown'
 
+        try:
+            self._ce = loadSiteLocalConfig().siteName
+        except SiteConfigError:
+            logger.error("can't load siteconfig, defaulting to hostname")
+            self._ce = socket.gethostname()
+
     def __del__(self):
         self.free()
 
@@ -94,6 +99,13 @@ class Monitor(DummyMonitor):
 
     def send(self, taskid, params):
         apmonSend(self._workflowid, taskid, params, logging, conf)
+
+    def generate_ids(self, taskid):
+        seid = 'https://{}/{}'.format(self._ce, sha1(self._workflowid).hexdigest()[-16:])
+        monitorid = '{0}_{1}/{0}'.format(taskid, seid)
+        syncid = 'https://{}//{}//12345.{}'.format(self._ce, self._workflowid, taskid)
+
+        return monitorid, syncid
 
     def register_run(self):
         self.send('TaskMeta', {
@@ -125,7 +137,7 @@ class Monitor(DummyMonitor):
             'broker': 'condor',
             'bossId': str(id),
             'SubmissionType': 'Direct',
-            'TargetSE': 'ndcms.crc.nd.edu',
+            'TargetSE': 'Many_Sites', # XXX This should be the SE where input data is stored
             'localId' : '',
             'tool': 'lobster',
             'JSToolVersion': '3.2.1',
@@ -153,7 +165,7 @@ class Monitor(DummyMonitor):
             'StatusValue': status,
             'StatusEnterTime':
             "{0:%F_%T}".format(datetime.datetime.utcnow()),
-            'StatusDestination': 'ndcms.crc.nd.edu',
+            'StatusDestination': self._ce,
             'RBname': 'condor'
             })
 
--- a/lobster/core/config.py
+++ b/lobster/core/config.py
@@ -88,7 +88,7 @@ class Config(Configurable):
         self.workflows = Items(workflows, key=lambda w: w.label)
         self.advanced = advanced if advanced else AdvancedOptions()
 
-        cats = list(set([w.category for w in workflows])) + [Category(name='merge', cores=1, memory=900)]
+        cats = list(set([w.category for w in workflows])) + [Category(name='merge', cores=1)]
         self.categories = Items(cats, key=lambda c: c.name)
 
         self.base_directory = base_directory
--- a/lobster/core/data/siteconfig/JobConfig/site-local-config.xml
+++ b/lobster/core/data/siteconfig/JobConfig/site-local-config.xml
@@ -1,5 +1,5 @@
  <site-local-config>
- <site name="T3_OSG_Opportunistic">
+ <site name="T3_US_Lobster">
     <event-data>
       <catalog url="trivialcatalog_file:siteconfig/PhEDEx/storage.xml?protocol=xrootd"/>
     </event-data>
--- a/lobster/core/data/siteconfig/JobConfig/site-local-config.xml
+++ b/lobster/core/data/siteconfig/JobConfig/site-local-config.xml
@@ -1,11 +1,11 @@
  <site-local-config>
- <site name="T3_US_NotreDame">
+ <site name="T3_OSG_Opportunistic">
     <event-data>
       <catalog url="trivialcatalog_file:siteconfig/PhEDEx/storage.xml?protocol=xrootd"/>
     </event-data>
     <calib-data>
      <frontier-connect>
-       <proxy url="http://eddie.crc.nd.edu:3128"/>
+       <proxy url="$HTTP_PROXY"/>
        <backupproxy url="http://cmsbproxy.fnal.gov:3128"/>
        <backupproxy url="http://cmsbpfrontier.cern.ch:3128"/>
        <server url="http://cmsfrontier.cern.ch:8000/FrontierInt"/>
--- a/lobster/core/data/task.py
+++ b/lobster/core/data/task.py
@@ -11,6 +11,7 @@ import re
 import resource
 import shlex
 import shutil
+import socket
 import subprocess
 import sys
 import traceback
@@ -18,8 +19,9 @@ import traceback
 sys.path.append('python')
 
 from WMCore.DataStructs.LumiList import LumiList
-from WMCore.Services.Dashboard.DashboardAPI import apmonSend, apmonFree
 from WMCore.FwkJobReport.Report import Report
+from WMCore.Services.Dashboard.DashboardAPI import apmonSend, apmonFree
+from WMCore.Storage.SiteLocalConfig import loadSiteLocalConfig, SiteConfigError
 
 import ROOT
 
@@ -704,8 +706,25 @@ if prologue and len(prologue) > 0:
 
 data['task timing']['prologue end'] = int(datetime.now().strftime('%s'))
 
-sync_ce = os.environ.get("LOBSTER_SYNC_CE", config['default ce'])
+try:
+    if os.environ.get("PARROT_ENABLED", "FALSE") == "TRUE":
+        raise ValueError()
+    sync_ce = loadSiteLocalConfig().siteName
+except Exception as e:
+    for envvar in ["GLIDEIN_Gatekeeper", "OSG_HOSTNAME", "CONDORCE_COLLECTOR_HOST"]:
+        if envvar in os.environ:
+            sync_ce = os.environ[envvar]
+            break
+    else:
+        host = socket.gethostname()
+        sync_ce = config['default host']
+        if host.rsplit('.')[-2:] == sync_ce.rsplit('.')[-2:]:
+            sync_ce = config['default ce']
+        else:
+            sync_ce = 'Unknown'
 target_se = config['default se']
+
+print ">>> using sync CE", sync_ce
 
 parameters = {
             'ExeStart': str(config['executable']),
--- a/lobster/core/data/task.py
+++ b/lobster/core/data/task.py
@@ -376,6 +376,9 @@ def copy_outputs(data, config, env):
     outsize = 0
     outsize_bare = 0
 
+    server_re = re.compile("[a-zA-Z]+://([a-zA-Z0-9:.\-]+)/")
+    target_se = []
+
     transferred = []
     for localname, remotename in config['output files']:
         # prevent stageout of data for failed tasks
@@ -415,6 +418,7 @@ def copy_outputs(data, config, env):
                         shutil.copy2(localname, rn)
                         if check_output(config, localname, remotename):
                             transferred.append(localname)
+                            target_se.append(default_se)
                             break
                     except Exception as e:
                         print e
@@ -444,6 +448,9 @@ def copy_outputs(data, config, env):
                 p = run_subprocess(args, env=pruned_env)
                 if p.returncode == 0 and check_output(config, localname, remotename):
                     transferred.append(localname)
+                    match = server_re.match(args[-1])
+                    if match:
+                        target_se.append(match.group(1))
                     break
             elif output.startswith("chirp://"):
                 server, path = re.match("chirp://([a-zA-Z0-9:.\-]+)/(.*)", output).groups()
@@ -459,6 +466,9 @@ def copy_outputs(data, config, env):
                 p = run_subprocess(args, env=env)
                 if p.returncode == 0 and check_output(config, localname, remotename):
                     transferred.append(localname)
+                    match = server_re.match(args[-1])
+                    if match:
+                        target_se.append(match.group(1))
                     break
             else:
                 print '>>> skipping unhandled stage-out method: {0}'.format(output)
@@ -468,6 +478,10 @@ def copy_outputs(data, config, env):
 
     data['output size'] = outsize
     data['output bare size'] = outsize_bare
+
+    if len(target_se) > 0:
+        return max(((se, target_se.count(se)) for se in set(target_se)), key=lambda (x, y): y)[0]
+    return default_se
 
 def edit_process_source(pset, config):
     """Edit parameter set for task.
@@ -690,10 +704,13 @@ if prologue and len(prologue) > 0:
 
 data['task timing']['prologue end'] = int(datetime.now().strftime('%s'))
 
+sync_ce = os.environ.get("LOBSTER_SYNC_CE", config['default ce'])
+target_se = config['default se']
+
 parameters = {
             'ExeStart': str(config['executable']),
             'NCores': config.get('cores', 1),
-            'SyncCE': 'ndcms.crc.nd.edu',
+            'SyncCE': sync_ce,
             'SyncGridJobId': syncid,
             'WNHostName': os.environ.get('HOSTNAME', '')
             }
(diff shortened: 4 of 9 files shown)


matz-e:
Pinging @annawoodard and @klannon, @khurtado for comments. There is still a way to go for this to be more general and figure out the right dashboard reporting à la CRAB/parsing job ads and environment to figure out the CE on the worker.

lobster/cmssw/dash.py
```diff
@@ -45,10 +47,7 @@ def __init__(self, workdir):
         self._workflowid = util.checkpoint(workdir, 'id')
 
     def generate_ids(self, taskid):
-        monitorid = '{0}_{1}/{0}'.format(taskid, 'https://ndcms.crc.nd.edu/{0}'.format(sha1(self._workflowid).hexdigest()[-16:]))
-        syncid = 'https://ndcms.crc.nd.edu//{0}//12345.{1}'.format(self._workflowid, taskid)
-
-        return monitorid, syncid
+        return str(taskid), str(taskid)
```
klannon: Just trying to understand what was done here.  It seems like this is treating monitorid and syncid as dummies and just returning the task ID for both.  Is this a placeholder or is that really how you'd want to treat it?  I suppose @khurtado should comment.

klannon: Wait, never mind.  I see it's just moved to another place.  Should read the whole thing first...

lobster/core/data/siteconfig/JobConfig/site-local-config.xml
```diff
@@ -1,11 +1,11 @@
  <site-local-config>
- <site name="T3_US_NotreDame">
+ <site name="T3_OSG_Opportunistic">
```
klannon: How was this sitename chosen?  T3_OSG_Opportunistic is actually being used to represent OSG (non-CMS) sites that we use.  Does the name chosen here get used anywhere else?  Does it have to correspond to something in SiteDB?  Could we choose something unique like T3_US_Lobster?  Again, I think @khurtado could comment...

lobster/cmssw/dash.py
khurtado: I think this is fine, since this is `generate_ids` from the `DummyMonitor`, that function is then overridden in the actual Monitor class

klannon:
OK, I just tried to look through this stuff.  I didn't really find anything except maybe using T3_OSG_Opportunistic as the site name in the provided file.  Anyway, I think this is good so I'm happy to see it merged.  @annawoodard or @khurtado?

lobster/cmssw/dash.py
```diff
@@ -86,6 +85,12 @@ def __init__(self, workdir):
         else:
             self.__executable = 'Unknown'
 
+        try:
+            self._ce = loadSiteLocalConfig().siteName
+        except SiteConfigError:
+            logger.error("can't load siteconfig, defaulting to hostname")
+            self._ce = socket.gethostname()
+
```
khurtado: I think `self._ce` should be simply 'Unknown'. What you have done is fine if we assume Lobster will be running in a local cluster and you know where you are expecting your jobs to be run, but when you run lobster on e.g OSG/CMS-Connect, then you don't really know what CE will your worker node run at. For example, if we set a lobster master on earth, then you will see "T3_US_NotreDame' in all jobs until they reach an actual worker node and update the CE information via syncCE.

lobster/cmssw/dash.py
```diff
@@ -95,6 +100,13 @@ def free(self):
     def send(self, taskid, params):
         apmonSend(self._workflowid, taskid, params, logging, conf)
 
+    def generate_ids(self, taskid):
+        seid = 'https://{}/{}'.format(self._ce, sha1(self._workflowid).hexdigest()[-16:])
+        monitorid = '{0}_{1}/{0}'.format(taskid, seid)
+        syncid = 'https://{}//{}//12345.{}'.format(self._ce, self._workflowid, taskid)
+
+        return monitorid, syncid
+
```
khurtado: Maybe we should change `self._ce` by something else here. As I said, lobster can in principle run in many CEs via OSG/CMS-Connect. I would be okay by using the hostname of the machine where the lobster master is executed rathen than a CE name you are expecting to run at...

lobster/core/data/siteconfig/JobConfig/site-local-config.xml
matz-e: This is not used anywhere in our code on the worker node (the file is sent to the worker and not used locally) – just making clear that this is not the NotreDame config.

lobster/cmssw/dash.py
matz-e: So if we set it to `Unknown`, the job updates will fix the dashboard reporting when they report their stuff?

lobster/core/data/siteconfig/JobConfig/site-local-config.xml
klannon: Ha!  Then let's go full bore on commercial branding and call it something like T3_ND_Lobster!

lobster/cmssw/dash.py
matz-e: But do these ids get parsed anywhere on the dashboard? In my understanding, this is just a unique identifier for each task and the workflow as a whole, so the actual content shouldn't matter?  CRAB2 actually had a condor identifier in there, IIRC. I think `grid-control` just had the submission host in those strings, too…

lobster/cmssw/dash.py
```diff
@@ -153,7 +165,7 @@ def update_task(self, id, status):
             'StatusValue': status,
             'StatusEnterTime':
             "{0:%F_%T}".format(datetime.datetime.utcnow()),
-            'StatusDestination': 'ndcms.crc.nd.edu',
+            'StatusDestination': self._ce,
```
matz-e: @khurtado is this what would throw the dashboard off and report all jobs as belonging to ND if earth is the master and the OSG runs the workers?

khurtado: Yes, I have something similar in CMS-Connect as 'Unknown', so the Site name in dashboard is just that when jobs are idle rather than some CE jobs haven't run at yet. After they run sync_ce takes over and the Site name is updated.

lobster/core/data/task.py
```diff
@@ -690,10 +704,13 @@ def get_bare_size(filename):
 
 data['task timing']['prologue end'] = int(datetime.now().strftime('%s'))
 
+sync_ce = os.environ.get("LOBSTER_SYNC_CE", config['default ce'])
```
annawoodard: Where does `LOBSTER_SYNC_CE` get set? Shouldn't we try to read the CE on the worker in the same way the master does it ([here](https://github.com/matz-e/lobster/pull/298/files#diff-a59f3942a951b8ad72c2997fe7aa81c3R108))?

lobster/cmssw/dash.py
khurtado: That's correct, it's just an identifier, so it strictly doesn't matter whether you use the ce name of the local cluster, the hostname of the submission host, or something else for it like 'lobster'. You can see the strings in the dashboard though [1], and you actually can use that id to pull a specific job data from dashboard [2]. I don't think much people will really even notice the ID strings or use it like in [2], so it's not that important and it really doesn't affect the functionality at all, but I think self._ce is not the best option if we end up setting that to 'unknown' as default for example. Of course that wouldn't affect the dashboard reporting, but a string id starting with 'unknown' would be ugly :)

[1] http://dashb-cms-job.cern.ch/dashboard/request.py/taskmonitoring#action=taskJobs&usergridname=Kenyi%20Paolo%20Hurtado%20Anampa&taskmonid=cmsconnect_khurtado_tutorial03_d31688c9143ed19c&what=s

[2] dashb-cms-job.cern.ch/dashboard/request.py/detailView?schedulerJobId=https://login.uscms.org/9641500d870aa8c1/2

lobster/core/data/task.py
matz-e: Yeah… that's a part that's not yet implemented. I was thinking we can try grabbing that from the job ad or then env in the wrapper, if needed. Note that the master way will not work within parrot, as we provide the inputs then…

lobster/cmssw/dash.py
matz-e: OK, `Unknown` would be bad. I would suggest leaving it like it is then, as the submission host seems to be a good fit.

lobster/cmssw/dash.py
matz-e: Then I will change/hardcode it to `Unknown`, unless someone beats me to it.

lobster/core/data/task.py
matz-e: (the environment would be a good way to grab it for generic non-CMS hosts, that are in OSG)

lobster/core/data/siteconfig/JobConfig/site-local-config.xml
khurtado: Wait, doesn't [1]  read the Site Name from `site-local-config.xml`?

[1] https://github.com/matz-e/lobster/blob/remove-hardcoded-ndcms/lobster/cmssw/dash.py#L89

lobster/core/data/task.py
khurtado: I got lost here. Where is `LOBSTER_SYNC_CE` set? Also, config['default ce'] would point you to the site name (or just a hostname when ` loadSiteLocalConfig().siteName` fails), not the CE fqdn. Does the dashboard accept both kinds of patterns for the sync_ce? I assumed it was only fqdn, but I might be wrong.

If it accepts both, great! If it doesn't and we need the CE fqdn maybe we can use WMCore again and query SiteDB via `SiteDBJSON::cmsNametoCE` to tell us the CE name. Edit: Actually... it seems that was 'disabled', although the fqdn from site_resources (which in principle should be the SE FQDNs)  are used to map from CE to SiteNames in [1]. Do you know if there is a WMCore tool/utility to get the CE FQDN based on the Site Name or if `cmsNametoSE` is ok considering that's what they do in[1]?

[1] https://github.com/dmwm/WMCore
/blob/8d324f4973bdf2a1b5eab2337b7938502107112c/src/python/WMCore/Services/SiteDB/SiteDB.py#L216-L230

annawoodard: @khurtado: I think you need to reload to pick up @matz-e's comment above:
> Yeah… that's a part that's not yet implemented. I was thinking we can try grabbing that from the job ad or then env in the wrapper, if needed. Note that the master way will not work within parrot, as we provide the inputs then…

khurtado: @annawoodard Ah, thanks! I should have reloaded before commenting. I still think the CE is the CE FQDN rather than the Site name though... when jobs are running through GlideinWMS, this is easy... as it can be grabbed from $GLIDEIN_Gatekeeper. Maybe I can take a look at the GlideInWMS framework to see how and when they set that for the WN.

annawoodard: > it can be grabbed from $GLIDEIN_Gatekeeper. Maybe I can take a look at the GlideInWMS framework to see how and when they set that for the WN.

Great!

I have a more even basic question for you, @khurtado or @matz-e: In case that isn't set, and assuming that we can get the CE from the CMS name (I'd say we try `SiteDBJSON::cmsNametoCE` as a first pass and only fix it if it's broken), what do you think is the most reliable way to get the CMS name on the worker? I would have assumed we should attempt `loadSiteLocalConfig()` on the worker, in case we're at a CMS site, before looking elsewhere?

annawoodard: And maybe set defaults via the `site-local-config` that we mount via Parrot?

annawoodard:
Could we use this to get the FQDN? 

[https://docs.python.org/2/library/socket.html#socket.getfqdn](https://docs.python.org/2/library/socket.html#socket.getfqdn)

lobster/core/data/task.py
annawoodard: Could we use this to get the FQDN? 

[https://docs.python.org/2/library/socket.html#socket.getfqdn](https://docs.python.org/2/library/socket.html#socket.getfqdn)

khurtado: @annawoodard  socket:getfqdn  would give us the FQDN from the WN, not the CE. As for the Site Name, I think pulling the site-local-config.xml from the worker nodes that are part of CMS Sites is the most reliable way indeed.

matz-e: One of my thoughts with the `default ce` was that you could check by domain name of the worker and keep it if it overlaps…

annawoodard: So, the procedure would be something like:

1. If `PARROT_ENABLED` is false, check the site-local-config for the site name, and use `SiteDBJSON::cmsNametoCE ` to get the CE from that.
2. Otherwise, check for `GLIDEIN_Gatekeeper`, `OSG_HOSTNAME`, or `CONDORCE_COLLECTOR_HOST`.
3. Otherwise, enforce some kind of mapping/overlap check between the worker name and the `default ce`.

matz-e: Sounds good!

lobster/core/data/siteconfig/JobConfig/site-local-config.xml
matz-e: That is on the master, but the local site-config is sent to the worker only.

lobster/cmssw/dash.py
matz-e: Will that really work, though? I just noticed that we call this function to update the status of a task from the master all the time… even after the task is done does not send further updates. Maybe we need to take that out then after the initial status update?

lobster/core/data/task.py
matz-e: @khurtado: Just noticed that `cmsNametoCE` will throw an [exception](https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/SiteDB/SiteDB.py#L140)?

matz-e: OK, re-read your comments. I'm running now without the site translation to CE to see what happens.

lobster/cmssw/dash.py
khurtado: Yeah, but my understanding is that `StatusDestination` won't override  the SiteName that easily. There are attributes that exist in Master and Worker nodes that can act on a same thing. In such cases, the WN attributes have priority. So, in this case... the WN `sync_ce` has priority over the Master-side `StatusDestination` and the latest updates the SiteName in cases where the dashboard didn't get the WN attribute or e.g the job failed and was resubmitted, as far as I remember at least...

lobster/core/data/task.py
khurtado: Oh right! Hope that works... not having to translate SiteName to the FQDN gatekeeper would simplify things. `SiteDBJSON::cmsNametoCE` does throw an exception, but the weird thing is that `SiteDBJSON::cmsNametoSE` and `SiteDBJSon::ceToCMSName` both use the SE FQDN for mapping... the latest just compares a string without verifying the FQDN type. Either that is a bug or Sites have SE and CE pointing to the same machine. I will ask WMCore guys...

lobster/cmssw/dash.py
matz-e: I can't currently get it to work, I'll run with some more debug output.

lobster/core/data/task.py
matz-e: I'm currently using the SE translation for now. At least for ND, they should be identical.

khurtado: WMCore guys replied back pretty fast. Short answer is: They don't support SiteName->CE or viceversa anymore.

matz-e: Do they have any alternative suggestions how to get the FQDN for the CE for a SiteName? Or what do they do to report things to the dashboard?

khurtado: No real alternatives from the WMCore side for that sort of mapping, they are also removing the CEs from SiteDB because they can't find someone that cares enough to maintain that info and only GlideinWMS uses the CEs, which is reflected in $GLIDEIN_Gatekeeper.

Good news is, they don't really seem to use the mapping in their dashboard, just the site name when available [here](https://github.com/dmwm/WMCore/blob/6f3570a741779d209f0f720647642d51b64845da/src/python/WMCore/WMRuntime/DashboardInterface.py#L18-L57). So they use the local site config, then some environment variables and the worker node host name in the worst case. We can follow a similar logic and forget about SiteName->CE mapping since that seems to work for them, but also keeping in mind we are able to set a master on a non-grid enabled resource without a local site config attached to it and that WNs will probably don't have the NORDUGrid or even the Globus GRAM environment variables, but can have those 3 others I mentioned.

matz-e: OK, thanks!  I had that in originally, I'll revert to it. I have a few things using the environment variables, I'll push them when I'm done merging fixes into previous commits.

matz-e:
OK, new take on things! I've tested it and after a the tasks start running, they correctly show up under `US_T3_NotreDame`…

matz-e:
I still have a lot of jobs in some weird `unknown` status, which we did not have before, like [here](https://dashb-cms-job-task.cern.ch/dashboard/templates/task-analysis/#user=Matthias+Wolf&refresh=300&table=Jobs&p=1&records=25&activemenu=0&status=&site=&tid=lobster_shuffle_3a60e74ed5b3259d). Any ideas, @khurtado? If this is solved, I think our dashboard reporting should be in a much better state…

khurtado:
Hi Mathias, woo! progress! Good thing SiteName works as the CE. So, I just looked at your task and it looks like the [final WN reporting](https://github.com/matz-e/lobster/blob/remove-hardcoded-ndcms/lobster/core/data/task.py#L877-L895) didn't go through for some reason. You can compare the parameters in [this](http://dashb-cms-job.cern.ch/dashboard/request.py/detailView?schedulerJobId=https://T3_US_NotreDame/397cf85893b3a81b/33) unknown job to the parameters in [this](http://dashb-cms-job.cern.ch/dashboard/request.py/detailView?schedulerJobId=https://T3_US_NotreDame/397cf85893b3a81b/11) successful job. You will notice the Executable timestamp, StageoutSE, Job Efficiency, etc are missing in the unknown one.

Could you activate dashboard debug mode for the WN? That should tell us what was reported and if dashboard complained about anything in particular while reporting some specific parameter...
```
import DashboardAPI
DashboardAPI.apmonLoggingLevel = "DEBUG"
```

Another thing that I noticed is that the WNHostname for successful jobs in your task is always `crcfe01.crc.nd.edu` , although the WN IP seems to be the actual IP address of the Worker Node rather than the IP of crcfe01. That's very strange... I think we only report the WN $HOSTNAME, don't we? 

```
Name:	crcfe01.crc.nd.edu
Address: 129.74.85.211
```

matz-e:
@khurtado turns out that it looks like the jobs in the unknown status seem to have failed due to `task.py` being temporarily changed by me messing around with git…

@annawoodard sounds good, although I'm not sure about the implications, and if the dashboard would even accept that – @khurtado? At least for FNAL, we shouldn't have to do that…

khurtado:
@matz-e: Oh right, great! Glad that is solved.

@annawoodard, @matz-e : 
So, WMCore passes the worker node hostname in the worst case scenario and let Dashboard handle the parameter as it pleases, we can do the same.

In general, I don't I think using either the full WN hostname or just the common domain part is that useful though (but I don't think it hurts in any way either). You can keep reading if you want to read some curious details about Dashboard :) 

[Here](http://dashb-cms-job-task.cern.ch/dashboard/request.py/taskmonitoring#action=taskJobs&usergridname=undefined&taskmonid=cmsconnect_khurtado_tutorial03_2ad7a0aaae6e7aa7&what=all) is a test where I forced `SyncCE = fnal.gov` and you can see SiteName was just `Unknown` in the Task Monitoring (same thing if you try to use SiteName filters with e.g Interactive View or Historical View), even though if you [query a specific taskID](http://dashb-cms-job.cern.ch/dashboard/request.py/detailView?schedulerJobId=https://login.uscms.org/029c5215b6adbbe1/1) you will see `fnal.gov` was passed.

You would think passing the full hostname might give Dashboard a chance to trace that back to some CMS Site (they have some code where they try to do that, according to Julia), but they also use the WNHostname for that. [Here](http://dashb-cms-job.cern.ch/dashboard/request.py/detailView?schedulerJobId=https://login.uscms.org/05c945485e552cdc/1) is another example where I forced "T3_US_Lobster" but Dashboard deemed that not to be a known CMS Site and preferred to trace back the WN to Wisconsin ("GLOW" is Wisconsin's SiteDB LCG SiteName).  Then, [here](http://dashb-cms-job.cern.ch/dashboard/request.py/detailView?schedulerJobId=https://login.uscms.org/f8d36965b76e9a41/0) is another case where I forced "T3_US_NotreDame" and dashboard felt happy enough with the name, so it didn't trace back the WN Hostname.

Also, keep in mind the worker node hostnames don't always have the `myhostname.domain.com` pattern. Meaning we would sometimes be delivering some random string (but again, Dashboard will most likely just ignore those cases). Here is one example of flat hostnames from OSG nodes when I ran lobster on it...
```
=hostname= abc-k20
=hostname= mwt2-c028
=hostname= abc-k22
=hostname= abc-k17
```

annawoodard:
@khurtado: Thanks for digging in here and discovering that the sitename will just be set to Unknown. I think that if we can see the syncCE when querying a specific task, we might as well save whatever info we're allowed to.

> Also, keep in mind the worker node hostnames don't always have the myhostname.domain.com pattern. 

@khurtado, @matz-e: Would using `socket.getfqdn()` be more robust then, and we can save the domain rather than just the hostname in the worst case scenario?

khurtado:
@annawoodard: Yes, I think `socket.getfqdn()` would definitely be more robust.  Using just the domain is fine with me too. I like the idea. @matz-e, thoughts?

lobster/core/config.py
```diff
@@ -88,7 +88,7 @@ def __init__(self, label, workdir, storage, workflows, advanced=None, plotdir=No
         self.workflows = Items(workflows, key=lambda w: w.label)
         self.advanced = advanced if advanced else AdvancedOptions()
 
-        cats = list(set([w.category for w in workflows])) + [Category(name='merge', cores=1, memory=900)]
+        cats = list(set([w.category for w in workflows])) + [Category(name='merge', cores=1)]
```
annawoodard: @matz-e: Just for my education, why is this needed? Do we know how much memory WQ will ask for?

annawoodard:
Looks great! Thanks for doing all this, @matz-e, and for all your testing, @khurtado!

lobster/core/config.py
matz-e: If I understand @btovar right, this will have the `merge` category consume all memory on a worker node until WQ figures out what the memory requirements are. I had issues with merge-jobs exceeding the memory requirements… (this was while testing the PR, hence it ended up in it)

matz-e:
Thanks for the thorough review, everyone!

lobster/core/config.py
annawoodard: Thanks!

PR #301, opened 2016-03-25
Remove yaml config support; Provide MC chained configuration.

matz-e:
This removes support for the YAML conversion shenanigans, and adds an MC configuration. As long as this pull request is not merged, I'll try to make the MC workflow work as well as possible and push any fixes. This should not keep anyone from merging the PR.

Fixes #288.

PR #303, opened 2016-03-29
Remove yaml config

annawoodard:
Here are a few final bits that were remaining of the yaml configurations. Sorry I missed this when I reviewed #301.

This leaves an MC example and a slimming example. I think this is sufficient example-wise and any further clarification belongs in the docs.


matz-e:
Excellent! I was hesitant to remove the other configs without a python equivalent or more feedback (should have mentioned that).

PR #307, opened 2016-03-31
MC generation tuning

matz-e:
Includes many fixes to make MC production work…

Tunes how merge tasks are created (as soon as enough work is done), and fixes the parent workflow handling. Bonus fixes mixed in!


annawoodard:
I have [tested](http://www.crc.nd.edu/~awoodard/lobster/crustacean_v1/) this branch with the new example config and everything looks... beautiful. :joy: Thanks, @matz-e!

PR #311, opened 2016-04-08
Fix unique arguments

annawoodard:
@matz-e: I'm pretty sure that moving `unique_arguments` from the workflow to the dataset specification is the way to go here. Unfortunately it required a lot of redundant copying around. Let me know if you think of a more elegant way to do it and I'll update the branch.

### DIFF
--- a/lobster/core/unit.py
+++ b/lobster/core/unit.py
@@ -364,8 +364,18 @@ class UnitStore:
                     False))
 
             for id, file, run, lumi, arg, failed in rows:
-                if (run, lumi) in all_lumis or failed > self.config.advanced.threshold_for_failure:
-                    logger.debug("skipping run {}, lumi {} with failure count {}".format(run, lumi, failed))
+                if (run, lumi) in all_lumis:
+                    logger.debug("skipping duplicate run {}, lumi {}".format(run, lumi))
+                    continue
+
+                if failed > self.config.advanced.threshold_for_failure:
+                    logger.debug("skipping run {}, "\
+                        "lumi {} "\
+                        "with failure count {} "\
+                        "exceeding `config.advanced.threshold_for_failure={}`".format(
+                            run, lumi, failed, self.config.advanced.threshold_for_failure
+                        )
+                    )
                     continue
 
                 if current_size == 0:
--- a/lobster/core/unit.py
+++ b/lobster/core/unit.py
@@ -364,7 +364,7 @@ class UnitStore:
                     False))
 
             for id, file, run, lumi, arg, failed in rows:
-                if (run, lumi) in all_lumis:
+                if (run, lumi, arg) in all_lumis:
                     logger.debug("skipping duplicate run {}, lumi {}".format(run, lumi))
                     continue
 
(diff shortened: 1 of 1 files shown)


matz-e:
Hm… I've got to say, this seems to make it more complex, and I'm not a fan of it. Rather than copying it into every dataset class, I think it would be better to keep it as a parameter of the workflow or have one dataset class that deals with the use case? OTOH, the latter may not be possible…

The whole treatment of unique arguments could use some re{working,thinking}, I think. I don't think that they work when using dependencies between datasets, since the 1-to-many mapping they introduce in terms of units between parent and child dataset makes accounting more complex than it could be.

annawoodard:
@matz-e: Ok, I've put it back `unique_arguments` the workflow. Now the only change introduced is the fix for skipping units.

> I don't think that they work when using dependencies between datasets, since the 1-to-many mapping they introduce in terms of units between parent and child dataset makes accounting more complex than it could be.

I think it is working ok, but I'll double check and open an issue if not.

matz-e:
Come to think of it, we should maybe change the name to `parametrization` instead of `unique_args` – and not just have one argument per unit but a generalized set of arbitrary parameters per unit? I'm very certain that if a child dataset has `unique_args` that things will break.

To fully support the use case, I think that we should group the sql query for units by `args` when creating tasks, so that we can have tasks with several units with the same argument?

annawoodard:
> Come to think of it, we should maybe change the name to parametrization instead of unique_args – and not just have one argument per unit but a generalized set of arbitrary parameters per unit? 

I really like that name! Are you suggesting we do the combinatorics for the user, or that we let the user provide a list of lists, with each inner list corresponding to one task? I would argue that since we've switched to a python config, we don't have much to gain by doing the combinatorics for the user-- it's more work for them to figure out what Lobster is going to do than it is to write a list comprehension, and it unnecessarily restricts them to certain combinatoric schemes. 

> I'm very certain that if a child dataset has unique_args that things will break.

Ah, I was only thinking about my own use case (where the first step has unique_args and nothing following it), which works. I think it would be nice to get this working for child datasets but until we have  any use cases for it, it should be low priority.

matz-e:
> I would argue that since we've switched to a python config, we don't have much to gain by doing the combinatorics for the user

Agreed! I was just thinking about other use-cases of this that I had while using grid-control, and I did not just vary one argument, but a set of arguments, hence parametrization? Still, providing the permutations is best done on the user side. We should also allow to have some of the arguments in the output-format? Maybe this works in python: `"{base}-{0}-{1}.{ext}".format(args_for_unit, base=…, ext=…)`…

> Ah, I was only thinking about my own use case (where the first step has unique_args and nothing following it), which works.

For now we can actually throw an exception when creating the configuration if a workflow has a `ParentDataset` and `unique_arguments` defined. Since this can be done in the constructor, it should be immediate feedback. Ultimately, we should provide fully-fledged support, which should include multi-unit tasks with the same argument and child dataset support.

annawoodard:
:+1:

PR #314, opened 2016-04-08
Mc plotting updates

matz-e:
Updates the MC generation config to 76, and fixes two minor monitoring issues. More to come…

### DIFF
--- a/examples/mc_gen.py
+++ b/examples/mc_gen.py
@@ -1,7 +1,7 @@
 from lobster import cmssw
 from lobster.core import *
 
-version = '_take27'
+version = '_take29'
 
 storage = StorageConfiguration(
         output=[
@@ -34,8 +34,8 @@ lhe = Workflow(
 
 gs = Workflow(
         label='gs_step',
-        pset='mc_gen/HIG-RunIIWinter15GS-00301_1_cfg.py',
-        sandbox_release='mc_gen/CMSSW_7_1_16_patch2',
+        pset='mc_gen/HIG-RunIISummer15GS-00177_1_cfg.py',
+        sandbox_release='mc_gen/CMSSW_7_1_18',
         merge_size='500M',
         dataset=ParentDataset(
             parent=lhe,
@@ -43,7 +43,7 @@ gs = Workflow(
         ),
         category=Category(
             name='gs',
-            cores=1,
+            cores=2,
             memory=2000,
             runtime=45 * 60
         )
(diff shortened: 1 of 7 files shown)


annawoodard:
:clap: :clap: :clap: 
Awesome!

annawoodard:
@matz-e: Sorry-- I should have tested before merging-- but trying your new MC configuration, I see:

```
  File "/afs/crc.nd.edu/user/a/awoodard/lobster/lobster/core/source.py", line 182, in __init__
    wflow.setup(self.workdir, self.basedirs)
  File "/afs/crc.nd.edu/user/a/awoodard/lobster/lobster/core/workflow.py", line 332, in setup
    self.determine_outputs(basedirs)
  File "/afs/crc.nd.edu/user/a/awoodard/lobster/lobster/core/workflow.py", line 307, in determine_outputs
    source = imp.load_source('cms_config_source', self.pset, f)
  File "mc_gen/HIG-RunIIFall15MiniAODv2-00224_1_cfg.py", line 88, in <module>
    process.Flag_chargedHadronTrackResolutionFilter = cms.Path(process.chargedHadronTrackResolutionFilter)
AttributeError: 'Process' object has no attribute 'chargedHadronTrackResolutionFilter'
```

matz-e:
@annawoodard: weird, it runs fine for me: http://www.crc.nd.edu/~mwolf3/lobster/test_mc_take29/

The shell script may actually have silent das errors that may mis-configure some steps (ran into that a few times). We could compare resulting configs (mine are in `~mwolf3/work/lobster/examples/mc_gen`).

matz-e:
Hmm… assuming I found your pset, they seem to be identical.

matz-e:
Ooooh… Which release are you using originally before you source lobster? I'm using my ttH work area, which is `7_6_3`…

annawoodard:
Good catch, that is probably it! Maybe we should add a `cmsrel` and `cmsenv` at the top of the script so you can run it from any release and get the same result?

annawoodard:
I will test and report back.

matz-e:
I think the script(s) do the right thing… but what are you running Lobster with? Because it fails when trying to load a `7_6_3` script in the local Lobster instance.

annawoodard:
> I think the script(s) do the right thing…

Wouldn't it be better if the script wasn't sensitive to what release you source it from?

I think I ran it with `CMSSW_7_6_1` but I will double check.

matz-e:
The script is downloading a bunch of setup scripts, each creating and sourcing their own release before creating the config, I think.

The question is which release environment Lobster is running in, because ultimately, that determines what the configuration sees when we load it to determine the pset outputs.

PR #316, opened 2016-04-09
Fix total lumis

annawoodard:
I think this gives what we should expect, using @klannon's dataset as a test case:

```
>>> from lobster import cmssw
>>> dset = cmssw.Dataset(dataset='/SingleMuMinusFlatPt0p2To150/TTI2023Upg14D-PU140bx25_PH2_1K_FB_V3-v2/GEN-SIM-DIGI-RAW', events_per_task=1000)
>>> info = dset.get_info()
>>> info.total_lumis
1001
>>> info.unmasked_lumis
1001
>>> info.masked_lumis
0
>>> dset.total_units
1681
```

I think this is the distinction between lumis and units that we want to make-- lumis should be unique, units should be the smallest pieces of work we can do (so one split lumi across two files is two units).

### DIFF
--- a/lobster/cmssw/dataset.py
+++ b/lobster/cmssw/dataset.py
@@ -146,6 +146,6 @@ class Dataset(Configurable):
                         if not mask or ((run['run_num'], lumi) in unmasked_lumis):
                             result.files[fn].lumis.append((run['run_num'], lumi))
 
-        result.masked_lumis = result.unmasked_lumis - result.total_lumis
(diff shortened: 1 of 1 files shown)


annawoodard:
This still has an issue with the summary on the plotting page:
screen shot 2016-04-10 at 12 16 19 am

What is shown under `units` is actually `lumis`. I think that units are more useful and simpler to deal with, so in my opinion we should switch to only using units instead of keeping track of both. I'm closing this and I'll try that approach in a separate PR.

PR #318, opened 2016-04-10
Fix lumi counting

annawoodard:
This respects the distinction between lumis and units that we want to make-- lumis should be unique (run, lumi) pairs, units should be the smallest pieces of work we can do (so one split lumi across two files is two units).
Here's what we get on the master branch, looking at @klannon's dataset:
screen shot 2016-04-10 at 2 10 40 am
And here's what this PR introduces:
screen shot 2016-04-10 at 2 19 09 am

Note that this doesn't fix the missing events yet.

### DIFF
--- a/lobster/cmssw/dataset.py
+++ b/lobster/cmssw/dataset.py
@@ -135,17 +135,18 @@ class Dataset(Configurable):
         else:
             blocks = self.__apis[instance].listBlocks(dataset=dataset)
             if mask:
-                unmasked_units = LumiList(filename=mask)
+                unmasked_lumis = LumiList(filename=mask)
             for block in blocks:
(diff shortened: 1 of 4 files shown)


annawoodard:
Oh, no! Shame on me for submitting a PR without running the tests. I'll look into why these failed tomorrow.

matz-e:
That's why the tests are run automatically for you :)

annawoodard:
@matz-e: I went in a few circles here, but I think I ended up at the right place. And the tests are passing now, yay! (It was just the renaming `total_lumis` -> `total_units`, not any content of the test that needed to be updated.)

matz-e:
ども　ありがとうございます！

annawoodard:
どういたしまして @matz-e!

PR #320, opened 2016-04-11
Find missing events

### LINKED ISSUE #319: Find lost events
In @klannon's recent [processing](http://www.crc.nd.edu/~klannon/tracklet_single_mu_withPU_v01/all/) run, Lobster thought he was done at 100k events even though DAS thinks the dataset has 168k. We need to understand this discrepancy-- are we skipping events? It's also possible that DAS is wrong. This dataset [1] is a good test case because it has many split lumis:

```
sqlite> select count(*) from (select distinct file, run, lumi from units_minus);
1681
sqlite> select count(*) from (select distinct run, lumi from units_minus);
1001
```

[1] /SingleMuMinusFlatPt0p2To150/TTI2023Upg14D-PU140bx25_PH2_1K_FB_V3-v2/GEN-SIM-DIGI-RAW

annawoodard:
We currently do '[lumi-mending](https://github.com/matz-e/lobster/blob/master/lobster/core/unit.py#L395)' of split lumis, where we pick a (run, lumi) that hasn't been finished, then do a second query to find all of the files with that (run, lumi) in it and add those to the processing of the current task. To try and debug the missing events issue, I wrote a little [script](https://github.com/matz-e/lobster/blob/find-missing-events/test/count_events.py) to count the unique events in a dataset.

Looking at the original dataset, to double check this isn't a problem in DAS (I copied it locally for faster debugging):

```
~/lobster/test >python count_events.py /hadoop/store/user/awoodard/minus
found 168000 (100000 unique) run, lumi, event sets in 192 files in path /hadoop/store/user/awoodard/minus
```

Here is the [result](http://www.crc.nd.edu/~awoodard/lobster/simple_crustacean_7be2ff1_v1/) of running with the master branch:

```
~/lobster/test >python count_events.py /hadoop/store/user/awoodard/simple_crustacean_7be2ff1_v1/minus/
found 100000 (100000 unique) run, lumi, event sets in 174 files in path /hadoop/store/user/awoodard/simple_crustacean_7be2ff1_v1/minus/
```

Here is the [result](http://www.crc.nd.edu/~awoodard/lobster/simple_crustacean_ba3a0a6_v1/) of running with the lumi-mending removed (this branch):

```
(.lobster)[earth] ~/lobster/test >python count_events.py /hadoop/store/user/awoodard/simple_crustacean_ba3a0a6_v1/minus/
found 167300 (100000 unique) run, lumi, event sets in 403 files in path /hadoop/store/user/awoodard/simple_crustacean_ba3a0a6_v1/minus/
```

So in @klannon's dataset, there are a large number of duplicate run, lumi, event sets. I don't think we want to throw events away just because they have duplicate run, lumi, event sets (for MC, at least.) I can think of innocuous scenarios leading to duplicates where the events themselves are perfectly useful (as long as the random seed is not fixed.) What is the use case we are trying to protect against-- mistakes upstream that result in real duplication of data? I think we need at least a knob to turn off lumi mending, although my vote would be to remove it completely (which this PR does) in order to keep things simpler. @matz-e, @klannon: opinions?

Fixes #319.

### DIFF
--- a/lobster/core/unit.py
+++ b/lobster/core/unit.py
@@ -102,7 +102,8 @@ class UnitStore:
             pset_hash text default null,
             publish_label text,
             release text,
-            uuid text)""")
+            uuid text,
+            single_file_max)""")
         self.db.execute("""create table if not exists tasks(
             bytes_bare_output int default 0 not null,
             bytes_output int default 0 not null,
@@ -176,8 +177,9 @@ class UnitStore:
                        units,
                        units_masked,
                        units_left,
-                       events)
-                       values (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)""", (
+                       events,
+                       single_file_max)
+                       values (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)""", (
                            wflow.label,
                            label,
                            wflow.label,
@@ -192,7 +194,8 @@ class UnitStore:
                            dataset_info.total_units * len(unique_args),
                            dataset_info.masked_units,
                            dataset_info.total_units * len(unique_args),
-                           dataset_info.total_events))
+                           dataset_info.total_events,
+                           getattr(wflow, 'single_file_max', False)))
 
         self.db.execute("""create table if not exists files_{0}(
             id integer primary key autoincrement,
@@ -297,8 +300,8 @@ class UnitStore:
                 Factor to apply to the tasksize.
         """
         with self.db:
-            workflow_id, tasksize = self.db.execute(
-                    "select id, tasksize from workflows where label=?",
+            workflow_id, tasksize, single_file_max = self.db.execute(
+                    "select id, tasksize, single_file_max from workflows where label=?",
                     (workflow,)).fetchone()
 
             logger.debug(("creating {0} task(s) for workflow {1}:" +
@@ -335,6 +338,7 @@ class UnitStore:
                     select id, file, run, lumi, arg, failed
                     from units_{0}
                     where file in ({1}) and status not in (1, 2, 6, 7, 8)
+                    order by file
                     """.format(workflow, ', '.join('?' for _ in chunk)), chunk))
 
             logger.debug("creating tasks from {} files, {} units".format(len(files), len(rows)))
@@ -342,9 +346,6 @@ class UnitStore:
             # files and lumis for individual tasks
             files = set()
             units = []
-
-            # lumi veto to avoid duplicated processing
-            all_lumis = set()
 
             # task container and current task size
             tasks = []
@@ -364,11 +365,6 @@ class UnitStore:
                     False))
 
             for id, file, run, lumi, arg, failed in rows:
-                if (run, lumi, arg) in all_lumis:
-                    logger.debug("skipping already processed unit with "\
-                            "run {}, lumi {}, arg {}".format(run, lumi, arg))
-                    continue
-
                 if failed > self.config.advanced.threshold_for_failure:
                     logger.debug("skipping run {}, "\
                         "lumi {} "\
@@ -379,41 +375,22 @@ class UnitStore:
                     )
                     continue
 
-                if current_size == 0:
-                    if num <= 0:
-                        break
-
                 if failed == self.config.advanced.threshold_for_failure:
                     logger.debug("creating isolation task for run {}, lumi {} with failure count {}".format(run, lumi, failed))
                     insert_task([file], [(id, file, run, lumi)], arg)
                     continue
 
-                if lumi > 0:
-                    all_lumis.add((run, lumi, arg))
-                    if arg is not None:
-                        cursor = self.db.execute("""
-                            select id, file, run, lumi
-                            from units_{0}
-                            where
-                                run=? and lumi=? and arg=? and
-                                status not in (1, 2, 6, 7, 8) and
-                                failed < ?""".format(workflow),
-                                (run, lumi, arg, self.config.advanced.threshold_for_failure))
-                    else:
-                        cursor = self.db.execute("""
-                            select id, file, run, lumi
-                            from units_{0}
-                            where
-                                run=? and lumi=? and
-                                status not in (1, 2, 6, 7, 8) and
-                                failed < ?""".format(workflow),
-                                (run, lumi, self.config.advanced.threshold_for_failure))
-                    for (ls_id, ls_file, ls_run, ls_lumi) in cursor:
-                        units.append((ls_id, ls_file, ls_run, ls_lumi))
-                        files.add(ls_file)
-                else:
-                    units.append((id, file, run, lumi))
-                    files.add(file)
+                if single_file_max and (len(files) == 1) and (file not in files):
+                    insert_task(files, units, arg)
+
+                    files = set()
+                    units = []
+
+                    current_size = 0
+                    num -= 1
+
+                units.append((id, file, run, lumi))
+                files.add(file)
 
                 current_size += 1
 
@@ -425,6 +402,10 @@ class UnitStore:
 
                     current_size = 0
                     num -= 1
+
+                if current_size == 0:
+                    if num <= 0:
+                        break
 
             if current_size > 0:
                 insert_task(files, units, arg)
--- a/lobster/cmssw/dataset.py
+++ b/lobster/cmssw/dataset.py
@@ -79,6 +79,7 @@ class Dataset(Configurable):
         self.dbs_instance = dbs_instance
 
         self.total_units = 0
+        self.single_file_max = False
 
     def __get_mask(self, url):
         if not re.match(r'https?://', url):
@@ -122,6 +123,7 @@ class Dataset(Configurable):
         if infos is None:
             raise IOError('dataset {} contains no files'.format(dataset))
         result.total_events = sum([info['num_event'] for info in infos])
+        total_lumis = sum([info['num_lumi'] for info in infos])
 
         for info in self.__apis[instance].listFiles(dataset=dataset, detail=True):
             fn = info['logical_file_name']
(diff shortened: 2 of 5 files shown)


annawoodard:
Oops, sorry about that. I'll fix the tests tomorrow.

matz-e:
We had a very specific reason/corner-case of double-processing to put that check in:

* Assume that a dataset has split lumis
* We process two files in one task, each contains a certain `(run, lumi)`
* We take a unit with `(run, lumi)` from file one, but not from file two, because the task size is big enough
* Since CMSSW looks at the lumi mask without any file information, and processes both units
* The second unit will still be marked unprocessed in the database, since we did not intend for it to be processed

Really, I think these screwed up datasets may be better processed in file-based mode. In fact, I'd be happy to remove these checks as you have done here, but I think we should add a warning to the dataset class about split lumis and advise caution/refuse to use lumi-based splitting. I think with everything that has been done in 7xy, this should not occur anymore, but explicit checks would be better.
Alternatively, we can add a unit-based processing which stops on file-boundaries, and removes the corner-case (this is in WMCore as an option, not sure if the CRAB uses it).

That being said, I personally don't need that check, and hence really like this PR, because I'm done with everything released with `CMSSW < 7xy`, and I think the official chain is purely lumi-based by now. I think this will (a) simplify the code and (b) speed up job creation as we reduce the amount of queries. But I still think we should add a caveat warning…

Curious as to what @klannon's opinion is on this matter?

BTW, I think that the test look at exactly this corner-case from above and would have to be re-written better.

**TL;DR** (this normally goes at the bottom of posts, at least in the corners of the web where I hang out): Let's merge this and add a check for duplicate lumis to the `cmssw.Dataset` class. Then we should stop creating tasks at file-boundaries (taking only units from one file per task).

annawoodard:
Thanks for explaining the corner case so clearly, I remember now. (Funny how the development goes in circles.)

> Let's merge this and add a check for duplicate lumis to the cmssw.Dataset class. Then we should stop creating tasks at file-boundaries (taking only units from one file per task).

Are you saying we should always stop at file boundaries (simpler, my preference) or only if the check for duplicate lumis has failed? If we always stop at file boundaries, do we need a check (is there some further danger the user needs to know about)?

matz-e:
For the sake of completely adjustable tasks, I was thinking that stopping at file boundaries should only required when running with complex datasets. This was one of our main features (task size adjustability to match runtime specification), and I don't think we should give that up. Yes, it'll be a bit more complex (I think basically just vetoing to stop on encountering a new file), but I don't think any more complex than what we have currently. It would be easiest if we just required datasets with split lumis to be run on "file-based" units, but can potentially lead to very long processing times.

annawoodard:
> For the sake of completely adjustable tasks, I was thinking that stopping at file boundaries should only required when running with complex datasets. 

By complex datasets, you mean, datasets with split lumis? So we do a check if `total_lumis` != `total_units` in `cmssw.Dataset`, then maybe set a flag in the workflow table that we can grab [here](https://github.com/matz-e/lobster/blob/master/lobster/core/unit.py#L301) telling us whether or not to limit the fileinfo query to 1 file [here](https://github.com/matz-e/lobster/blob/master/lobster/core/unit.py#L318)?

> This was one of our main features (task size adjustability to match runtime specification), and I don't think we should give that up.

Good point! I was being too aggressive in my quest for simplicity.

annawoodard:
EDIT: That wasn't quite right because we wouldn't have to limit the query, but you get the idea.

matz-e:
> By complex datasets, you mean, datasets with split lumis? So we do a check if `total_lumis` != `total_units` in `cmssw.Dataset`, then maybe set a flag in the workflow table to pick up [here](https://github.com/matz-e/lobster/blob/master/lobster/core/unit.py#L301) telling us whether or not to stop on file boundaries?

Sounds like a plan. Just looked through the code, didn't you add `total_lumis` somewhere already?

annawoodard:
I traded `total_lumis` for `total_units`, but I can add it back in (we get it from the DBS API.) I'll go ahead and draft it up and add it to this branch for review. Do we have a preferred terminology for stopping on file boundaries (how do they call it in WMCore again?)

matz-e:
It's [`halt_job_on_file_boundaries`](https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/JobSplitting/EventAwareLumiBased.py#L45), which I think is a little bulky and not so clear. How about something like `single_file_mode` or `single_file_per_task`?

klannon:
Sorry for the radio silence.  I don't think I have any great technical insights.  My general suggestions of favoring simplicity over being able to address every possible permutation automatically apply here.  I am a little torn because I suspect that in this case, Lobster is actually saving me from having duplicated events because I think that there are really only supposed to be 100k events and when you "fix" Lobster and the output goes to 168k, many of those will be duplicates.  Nonetheless, I don't think this is a problem that Lobster should be solving.  In *many* cases (as @annawoodard points out), different events are accidentally generated with identical run and lumi section numbers.  In those cases Lobster would prevent users from accessing the full stats of the sample.

So, I support implementing whatever is the simplest approach.  If that approach would lead to not processing a full dataset in some weird corner cases, then we should have a clear alternative option or setting for dealing with that.  Quick question about file-based processing: if my files live somewhere else (say via xrootd) can I specify the list of files in a text file?  I'm only seem to remember being able to do this by pointing Lobster at a local directory.

matz-e:
@klannon, I think what @annawoodard brought up and what we're discussing here would lead to a simplification, without loosing the corner-cases. If your events are duplicated, you would have them in your output twice, though.

As you say, it's not really something Lobster should solve, and as long as you have EDM output, merging everything into one big file would get rid of duplicated `(run, lumi, event)` events.

Not sure about the file-based processing, there should be settings to use file-based processing with datasets from DBS. We should actually be able to do some xrootd query for datasets living elsewhere?

annawoodard:
> As you say, it's not really something Lobster should solve, and as long as you have EDM output, merging everything into one big file would get rid of duplicated (run, lumi, event) events.

Ah, thanks for reminding me about this. Now I'm going in circles because, depending on your merging parameters, you could have a many-to-one mapping between input and output events unless we break merging. Maybe it's better if we just leave things the way they are (except add in a warning about split lumis and warn that if there are duplicated `(run, lumi, event)`s that the 'events' might not equal 'read' in the Lobster summary.) If events MAY be deduplicated it's probably better to always deduplicate (as we're doing now) than sometimes deduplicate in a hard-to-predict way, since never deduplicating isn't an option.

klannon:
I guess the real question I have is, let's say right now I have some files out there somewhere (i.e. not local, but accessible via any of the usual mechanisms like xrootd, srm, or chirp) but they're not in DBS, is there a way to ask Lobster to process them file-by-file?  If not, then I'll make an issue requesting this mode.  If there is, then I think that plus the approach your discussing above handles every possible case, right?  And yes, I agree that if I've got duplication, that's my problem, not Lobster's.

Argh--too slow.  @annawoodard: I'm very confused about what you write.  I think we always want Lobster to assume it should process everything.  It should never try to detect and correct duplication in the input list.  Or am I missing something?

annawoodard:
I guess the real question I have is, let's say right now I have some files out there somewhere (i.e. not local, but accessible via any of the usual mechanisms like xrootd, srm, or chirp) but they're not in DBS, is there a way to ask Lobster to process them file-by-file?

[This](http://lobster.readthedocs.org/en/latest/config.html#lobster.core.dataset.Dataset) should work, it takes lists or directories.

> I think we always want Lobster to assume it should process everything. It should never try to detect and correct duplication in the input list. Or am I missing something?

My point is that, if we don't correct duplication (which we're doing now), than some duplication will be corrected if your merge size is big enough and the two duplicated events end up in the same file. It would be better to always correct, so that the output is predictable, rather than have some events de-duplicated randomly.

klannon:
Duplicate removal CMSSW is optional.  We should merge with the CMSSW 
parameter that doesn't remove duplicates.  We could allow duplicate 
removal as an advanced option.  Pure ROOT merging won't ever remove 
duplicates, so we should document the option clearly.

matz-e:
@annawoodard, @klannon to use the `core.Dataset` class with XrootD, you may have to specify the input parameters of the `StorageConfiguration` right.

lobster/core/unit.py
```diff
@@ -389,7 +392,7 @@ def insert_task(files, units, arg):
 
                 current_size += 1
 
-                if current_size == tasksize:
+                if (current_size == tasksize) or (single_file_max and len(files) > 1):
```
matz-e: Not sure your logic is correct here. If `len(files) > 1` already, don't we have two files in the task, but we want to have one only?

annawoodard: Thanks!

annawoodard:
@matz-e: This isn't as important as the scale test preparations, so I'll return to it and fix the tests afterwards.

annawoodard:
I do not feel as gung-ho as I did about this earlier. We have this in the history if we want to resurrect it later I don't think it's worth the danger of introducing new bugs right now.

matz-e:
This clearly needs a rebase — I think it would be worth it. If we have datasets that stop at the file boundary, we can have datasets with "continuous" lumi requirements — enabling us to enable LHE tasks that can process several units at once, and give us **dynamic LHE task sizes**. Code changes are fairly minor for that, too. I could quickly hack up something and hand it over to @WenzhaoLi for verification, testing, and approval?

annawoodard:
+1 sounds good to me

klannon:
+1 from me too, but only if your advisor doesn't object to your spending time on this instead of your thesis.  I definitely like the idea of not having you do it all, but letting @WenzhaoLi get involved.

WenzhaoLi:
Sorry for being late to the party. I will put this into my to-do-list.

matz-e:
I can rebase this for you and add some required changes… that should be a good start.

WenzhaoLi:
That'd be great! Thanks @matz-e !

PR #321, opened 2016-04-18
Remove bitwise comparison.

annawoodard:
This should be a direct comparison now, otherwise we're evaluating
true for `wq.WORK_QUEUE_RESULT_MAX_RETRIES` when the result is
`wq.WORK_QUEUE_RESULT_RESOURCE_EXHAUSTION`.


matz-e:
¡Gracias!

PR #322, opened 2016-04-18
Add knob to cleanup input files after processing.

### LINKED ISSUE #313: Add cleanup for intermediate files
In some projects with dependent workflows, the user may have intermediate outputs that he or she doesn't care about. These are impossible to cleanup by hand until the project is finished, as the user can't know which files are still needed for subsequent tasks. We should add a knob in the configuration which notifies the master these should be cleaned up as dependent tasks finish. This will be important for large projects in order to conserve space on the storage element.

annawoodard:
This removes the `cleanup_merged` knob for clarity-- I can't
think of any use cases for it. Note that this is a more
general implementation than only cleaning up intermediate
files for dependent datasets: users can delete any input
files as they are processed. Fixes #313.

### DIFF
--- a/lobster/core/source.py
+++ b/lobster/core/source.py
@@ -510,7 +510,7 @@ class TaskProvider(object):
                     for dep in wflow.dependents:
                         propagate[dep.label][outfn] = outinfo
 
-                if merge and wflow.merge_cleanup:
+                if merge or wflow.cleanup_input:
                     files = handler.input_files
                     cleanup += files
 
--- a/lobster/core/workflow.py
+++ b/lobster/core/workflow.py
@@ -95,8 +95,8 @@ class Workflow(Configurable):
             to.
         publish_label : str
             The label to be used for the publication database.
-        merge_cleanup : bool
-            Delete merged output files.
+        cleanup_input : bool
+            Delete input files after processing.
         merge_size : str
             Activates output file merging when set.  Accepts the suffixes
             *k*, *m*, *g* for kilobyte, megabyte, …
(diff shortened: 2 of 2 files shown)


matz-e:
If I read this right, merged input files will always be cleaned up, is that the intention?

annawoodard:
Yeah-- sorry if that wasn't clear. All merged input files will be cleaned up, and that's the intention-- I can't think of a case where you would want to keep them.

matz-e:
To give a counter-argument, you could have a scenario where we want to process dependent steps immediately (before merging), but also merge, e.g., the RECO step to be able to re-process it differently later. This needs advanced support at the moment, anyways, so let's merge this.

annawoodard:
That's a good point. However, when we get there, let's think again about whether or not that functionality is worth the added complication, because sometimes it may be better to keep things simple. If there are dependent datasets, might it be better to always process the dependent steps before merging?

klannon:
Right.  I'd argue you'd still want to clean up the unmerged files at the end.  Really, you'd just want to swap the order of merge and process dependent jobs.  (In fact, they could go in parallel, but that might make cleanup more complicated?)

matz-e:
It was easiest to implement the way it is, processing after merging. Otherwise, we need to keep track if a file is merged and processed, and I wanted to keep the inter-workflow interactions low.

matz-e:
For the record: As it is now, the only interaction is that after a regular processing/merging, the new units are inserted into the child database… the cleanup @annawoodard just implemented then deletes something completely processed/merged.

PR #326, opened 2016-04-19
More plotting fixes

### LINKED ISSUE #324: Avoid plotting zombies
Plotting processes are not exiting properly and are hanging around and taking up memory.

I have verified that this is happening on the master branch, and that it disappears when automatic plotting is disabled.

:skull: :no_entry_sign:

matz-e:
I think this is fairly stable, but may need a bit more tuning. I'm putting this out for review, to see if behavior w.r.t. threads has changed. In my recent run, Lobster did not spawn any additional processes, although the accounting was a little messed up:

http://www.crc.nd.edu/~mwolf3/lobster/test_mc_take38/

Fixes #324, hopefully.


annawoodard:
Do you mean, fixes #324?

matz-e:
Err… yes. Although the circular dependency was nice :)

annawoodard:
What is the problem with the accounting? I'm not seeing it at first glance.

This isn't related to this PR completely, but is there anything we can change to shorten the long tail here?

matz-e:
820 units merged for GS, but 972 processed for DIGI — that does not seem right. That's where the tail comes from, too. I'll look into it, although I have to be AFK again in a bit.

lobster/commands/process.py
```diff
@@ -94,6 +46,36 @@ def help(self):
...
+                        ["#timestamp", "total_create_time", "total_return_time", "units_left"] + self.log_attributes
+                    ) + "\n"
+            )
+
+    def log(self, category, creating, destroying, left):
+        filename = os.path.join(self.config.workdir, "lobster_stats_{}.log".format(category))
+        if category == 'all':
+            stats = self.queue.stats_hierarchy
+        else:
+            stats = self.queue.stats_category(category)
```
annawoodard: Do we know how this will behave if formen are being used? Does `queue.stats_category` take them into account as `stats_hierarchy` does?

matz-e: That's a question for @btovar, actually. I'd hope it will take it into account…

annawoodard: @btovar pointed out on Slack that the only field that matters is `workers_able`, and he needs to fix that field so that foremen are counted as the number of workers below them. So it looks like our treatment here is correct (it won't need to change after he fixes `stats_category`). :+1:

lobster/actions.py
```diff
@@ -82,6 +70,13 @@ def take(self, force=False):
 
         now = datetime.datetime.now()
         if (now - self.__last).seconds > 15 * 60 or force:
-            self.plotq.put('plot')
+            if not force and hasattr(self, 'p') and self.p.is_alive():
```
annawoodard: :+1:

annawoodard:
There's a lot of important updates here. I vote that, to reduce conflicts with other development that needs to happen before the scale run, we factorize out the accounting issue and merge this now as everything else looks great to me. Thanks, @matz-e! 👏 👏 👏

annawoodard:
Hmm, I'm still seeing some zombie processes with the merged master...

```
awoodard  885358  5.7  0.5 987452 673092 ?       S    19:02  17:09 /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/python /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/lobster process ttZ.py
awoodard 1124606  0.0  0.5 1214816 667844 ?      S    19:35   0:00 /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/python /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/lobster process ttZ.py
awoodard 1124773  0.0  0.5 987480 668308 ?       S    19:35   0:01 /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/python /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/lobster process ttZ.py
awoodard 1124775  0.0  0.5 987480 668740 ?       S    19:35   0:02 /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/python /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/lobster process ttZ.py
awoodard 1124776  0.0  0.5 987480 668700 ?       S    19:35   0:01 /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/python /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/lobster process ttZ.py
awoodard 1124778  0.0  0.5 987480 668660 ?       S    19:35   0:01 /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/python /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/lobster process ttZ.py
awoodard 1124825  0.0  0.5 987480 668824 ?       S    19:35   0:02 /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/python /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/lobster process ttZ.py
awoodard 1124829  0.0  0.5 987480 669112 ?       S    19:35   0:02 /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/python /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/lobster process ttZ.py
awoodard 1124831  0.0  0.5 987480 668752 ?       S    19:35   0:02 /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/python /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/lobster process ttZ.py
awoodard 1124837  0.0  0.5 987480 668856 ?       S    19:35   0:02 /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/python /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/lobster process ttZ.py
awoodard 1124839  0.0  0.5 987480 668412 ?       S    19:35   0:01 /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/python /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/lobster process ttZ.py
awoodard 1124842  0.0  0.5 987480 668456 ?       S    19:35   0:00 /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/python /afs/crc.nd.edu/user/a/awoodard/.lobster/bin/lobster process ttZ.py
```

matz-e:
That's weird. I haven't observed any additional processes besides one that turns `<defunct>` after finishing plotting. Which python version (from which release) are you using?

klannon:
One other random thought:  Could you be running Lobster in a different way such that @annawoodard's version runs into a problem that @matz-e's doesn't.  Here's an example:  It took me a long time to figure out that if you log out, the plotting stops working because your session loses its AFS token.  Probably you both know this.  However, if one of you (say @annawoodard) didn't know this but the other (say @matz-e) did, then presumably @matz-e would do something fancy like use screen to keep the AFS token fresh while @annawoodard might not.  Then @matz-e would never see what happens when the AFS token is lost.  (Maybe it causes a crash that generates zombie threads?)

Notes:  I picked the roles for @annawoodard and @matz-e above not because of my opinion about who knows more fancy computing but because it was convenient in supporting my hypothesis.  I expect, in reality that the AFS thing isn't the real explanation because you're both probably running `xscreen16 --autoAFS --never-crash --too-cool-for-Lannon`.  But maybe there's something else about your environment or session beyond Python, CMSSW, etc. versions that could explain the difference.

matz-e:
That may definitely be the case. I'm not sure about the AFS token… I don't use screen (at all, I hate it) or tmux (any more), and I don't mess with my permissions for the plotting mostly (I can do that, though, granting access to the plotting directory such that the lost AFS token doesn't matter anymore).

Another major difference could be the shell used (I'm on bash), but I think that in general, the CMSSW environment may well be factoring in, I've seen emails floating around on HN that make me **not trust** the people involved in providing an OS w/o kernel on top of RH (that's what CMSSW as evolved into) to mess around with very elementary parts of how stuff is run.</rant>

klannon:
Just a <b>test</b>.  Please ignore.

annawoodard:
I am using tmux (hah, the opposite of your guess!) Good brainstorming! But I'm not sure that tmux is helping me much with the AFS tokens, I still have to update them by hand every once in a while. I am using CMSSW_7_6_1, python 2.7.6.

> (I can do that, though, granting access to the plotting directory such that the lost AFS token doesn't matter anymore)

Ah, we should add a note about this to the documentation.

matz-e:
I was told not to use tmux (or a modern editor), so I don't know how to exactly reproduce this. I have another small fix where I'm wondering if it'll help… I'll push an update in a bit.

In my current workflows, I also don't rely on tmux being available (partially because I work on lxplus a lot, where the machine changes all the time), and I don't see a technical reason for running Lobster in tmux/screen.

matz-e:
FWIW, the wrong merge numbers were me using a bogus column (:facepalm:)

annawoodard:
> I was told not to use tmux (or a modern editor), so I don't know how to exactly reproduce this.

I'm not convinced it has anything to do with tmux. I'll try my next tests outside of tmux to confirm and report back.

> I don't see a technical reason for running Lobster in tmux/screen.

I don't use tmux specifically for Lobster, I just find it speeds up my workflow. This is especially true while I'm in Chicago and I have to sign off/reconnect quite frequently.

> FWIW, the wrong merge numbers were me using a bogus column (:facepalm:)

👏 👏 👏  Yay, you found it!

annawoodard:
Actually, to use the scientific method, I'll keep running Lobster in the same way and test with your changes first, to see if it was your change that fixed my problem. If I still have the problem I'll try outside of tmux.

PR #331, opened 2016-04-20
Add timeout for lcg-cp. Fixes #329.

### LINKED ISSUE #329: Fallback system breaks if SRM hangs
A common symptom of SRM unhealthiness is that it will hang. In this case, the fallback system breaks because we never move on to the next stageout method. We should add a timeout to avoid this.

annawoodard:


matz-e:
Awesome! :clap: :clap: :clap:

matz-e:
(this has bitten me quite a lot)

PR #337, opened 2016-04-24
Only delete files which are not needed anymore. Fixes #330.

### LINKED ISSUE #330: Do not try to run on files that don't exist
I haven't done any sleuthing yet, so I'm not sure where this is coming from, but I have quite a few [jobs](http://www.crc.nd.edu/~awoodard/lobster/ttZ/v2/a/logs/9539.log) that are failing because they are trying to run over files that don't exist, so it seems like something isn't getting updated correctly somewhere.

annawoodard:
Better tested-- this time I'm not throwing away work that we need!

Unfortunately, this requires an additional DB query in order to find out which files are still needed by other tasks. But the query will only be made if you've specified to cleanup inputs.

### DIFF
--- a/lobster/core/source.py
+++ b/lobster/core/source.py
@@ -483,6 +483,7 @@ class TaskProvider(object):
         cleanup = []
         update = defaultdict(list)
         propagate = defaultdict(dict)
+        input_files = defaultdict(set)
         summary = ReleaseSummary()
 
         for task in tasks:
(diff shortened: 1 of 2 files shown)


matz-e:
Looks great! To be clear, this will not yet allow us to process dependent steps before merging?

annawoodard:
That's correct.

PR #343, opened 2016-05-04
Add fallback summary

annawoodard:
This was a little tricky because there are can be different sets of protocols for each task, depending on the user's `StorageConfiguration`. I chose to save json strings to the SQL DB-- it's cramming a square peg in a round hole-- we should use a NoSQL DB to do this properly, or add all possible protocols as columns in the SQL DB. I chose this approach to avoid adding all of the new columns, even though it's a hack. I was worried about performance, so I'm only adding transfer info to the workflows table for now. If we find we need more granularity we can add it to the tasks table later.

I was a little unhappy with the nomenclature here, because I wanted each protocol in the table to map to its prefix for the `StorageConfiguration`, but `root` has two possible actions (xrdcp or streaming). I went with `xrdcp` for xrdcp and `root` for streaming.

I felt that a table was the clearest way to display this data, rather than multiple pie charts (I'm not convinced a pie chart is ever actually clearer than a table.) Output looks like this (stage-in is blank for this test because I am generating events).

screen shot 2016-05-04 at 3 47 26 pm

### DIFF
--- a/lobster/commands/data/category.html
+++ b/lobster/commands/data/category.html
@@ -184,6 +184,26 @@
             <a href="cpu-wall-hist.png"><img alt="" src="cpu-wall-hist.png"/></a>
             <a href="cpu-wall-int-hist.png"><img alt="" src="cpu-wall-int-hist.png"/></a>
         {% endif %}
+        {% if good_tasks or bad_tasks %}
+	    <h3>Transfer Protocol Summary</h3>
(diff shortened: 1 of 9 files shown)


annawoodard:
@matz-e: tests are fixed

matz-e:
Thanks! Nice use of the `Counter` object!

Issue #344, opened 2016-05-05
Implement `specify_category_autolabel_resource`


annawoodard:
I am separating this from #336 because I don't think it is necessary to have finished before the Summer 2016 scale test.

PR #345, opened 2016-05-05
Update category handling

### LINKED ISSUE #336: Update category handling with features from cctools 082-9ad9abb3-cvmfs-40cf5bba
See @btovar's summary on Slack below. We need to make sure we have the capability of taking advantage of these new features. The new quantities to be plotted are already part of #323.

```
##
    # Turn on or off first-allocation labeling for a given category. By
    # default, all resources are labeled, but wall, and cpu time. Turn on/off
    # specific resources with @ref specify_category_autolabel_resource.
    # NOTE: autolabeling is only meaningfull when task monitoring is enabled
    # (@ref enable_monitoring). When monitoring is enabled and a task exhausts
    # resources in a worker, mode dictates how work queue handles the
    # exhaustion:
    # @param self Reference to the current work queue object.
    # @param category A category name. If None, sets the mode by default for
    # newly created categories.
    # @param mode One of @ref category_mode_t:
    #                  - WORK_QUEUE_ALLOCATION_MODE_FIXED Task fails (default).
    #                  - WORK_QUEUE_ALLOCATION_MODE_MAX If maximum values are
    #                  specified for cores, memory, or disk (e.g. via @ref
    #                  specify_max_category_resources or @ref specify_memory),
    #                  and one of those resources is exceeded, the task fails.
    #                  Otherwise it is retried until a large enough worker
    #                  connects to the master, using the maximum values
    #                  specified, and the maximum values so far seen for
    #                  resources not specified. Use @ref specify_max_retries to
    #                  set a limit on the number of times work queue attemps
    #                  to complete the task.
    #                  - WORK_QUEUE_ALLOCATION_MODE_MIN_WASTE As above, but
    #                  work queue tries allocations to minimize resource waste.
    #                  - WORK_QUEUE_ALLOCATION_MODE_MAX_THROUGHPUT As above, but
    #                  work queue tries allocations to maximize throughput.
##
    # Turn on or off first-allocation labeling for a given category and
    # resource. This function should be use to fine-tune the defaults from @ref
    # specify_category_mode.
    # @param q A work queue object.
    # @param category A category name.
    # @param resource A resource name.
    # @param autolabel True/False for on/off.
    # @returns 1 if resource is valid, 0 otherwise.
In summary,   q.specify_category_mode(q, "my_cat",  WORK_QUEUE_ALLOCATION_MODE_FIXED) is the default. Tasks are NOT retried on any exhaustion, even on small workers.

WORK_QUEUE_ALLOCATION_MODE_MAX tries with at least the maximum seen, but in a worker that is larger than the maximum seen.

MIN_WASTE and MAX_THROUGHPUT try to be smart about what to allocate first.

Also, work_queue_status -A  mymasterhost mymasterport  shows a summary of workers that could run waiting tasks on a given category, together with the max resources seen.

The above is ready in  082-9ad9abb3-cvmfs-40cf5bba
```

annawoodard:
Let's consider merging this now, since we've opened #344 as a separate issue. Fixes #336.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -96,6 +96,6 @@ easy modification of the source::
 .. [#ftools] ``tcsh`` users should use the following to access the
    `cctools` development version at Notre Dame::
 
-    setenv cctools cctools-077-77be65c7-cvmfs-40cf5bba
+    setenv cctools cctools-083-7f811783-cvmfs-40cf5bba
     setenv PYTHONPATH ${PYTHONPATH}:/afs/crc.nd.edu/group/ccl/software/$cctools/x86_64/redhat6/lib/python2.6/site-packages
     setenv iATH /afs/crc.nd.edu/group/ccl/software/$cctools/x86_64/redhat6/bin:${PATH}
--- a/install_dependencies.sh
+++ b/install_dependencies.sh
@@ -23,8 +23,8 @@ fi
 
 (
 	cd $VIRTUAL_ENV/src
-	wget -O - http://ccl.cse.nd.edu/software/files/cctools-current-source.tar.gz|tar xzf -
-	cd cctools*source
+	wget -O - https://github.com/btovar/cctools/archive/ndcms_mid2016_scale_run.tar.gz|tar xzf -
+	cd cctools*
 	sed -i 's/\(config_perl_path\)=auto/\1=no/' ./configure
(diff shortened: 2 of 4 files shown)


matz-e:
Merging this as discussed in the meeting yesterday…

Issue #346, opened 2016-05-05
Generalize Chirp error handling


annawoodard:
From my recent master crash:

```
2016-05-05 06:49:11 [DEBUG] lobster.se: resolving file system method 'remove' with arguments (u'lhe_step/HIG-RunIIWinter15wmLHE-00196_5.root',), {}
Traceback (most recent call last):
  File "/afs/crc.nd.edu/user/a/awoodard/.lobster/lib/python2.7/site-packages/Lobster-1.5-py2.7.egg/lobster/./commands/process.py", line 320, in sprint
    task_src.release(tasks)
  File "/afs/crc.nd.edu/user/a/awoodard/.lobster/lib/python2.7/site-packages/Lobster-1.5-py2.7.egg/lobster/core/source.py", line 537, in release
    fs.remove(*cleanup)
  File "/afs/crc.nd.edu/user/a/awoodard/.lobster/lib/python2.7/site-packages/Lobster-1.5-py2.7.egg/lobster/se.py", line 70, in switch
    return imp.fixresult(getattr(imp, attr)(*map(imp.lfn2pfn, args), **kwargs))
  File "/afs/crc.nd.edu/user/a/awoodard/.lobster/lib/python2.7/site-packages/Lobster-1.5-py2.7.egg/lobster/se.py", line 261, in remove
    self.__c.rm(path)
  File "/afs/crc.nd.edu/group/ccl/software/cctools-099-282610d4-cvmfs-40cf5bba/x86_64/redhat6/lib/python2.6/site-packages/Chirp.py", line 1160, in rm
    status = chirp_reli_rmall(self.hostport, path, self.__stoptime(absolute_stop_time, timeout))
TypeError: in method 'chirp_reli_rmall', argument 2 of type 'char const *'
```

Let's try to generalize `OSError` and `IOError` so that we don't crash the master for errors like this.

btovar:
I can only reproduce this by passing a non-string argument as a path. Is there a chance of this happening in lobster?

klannon:
What happens if you pass in `None`?  That's the obvious candidate coming in from the python API.  Do we just need to be careful that what we pass isn't none?  Alternatively, we could `str()` anything we pass into Chirp...

btovar:
None throws an OSError, as it is interpreted as the string '', and there is not such a file.

btovar:
None throws an OSError, as it is interpreted as the string ''.

matz-e:
[...]

PR #347, opened 2016-05-05
Catch `TypeErrors` from Chirp.

annawoodard:
This hack addresses #346 so that we don't bring down the master
for this error, but let's leave the issue open until we can
solve it correctly.


matz-e:
I've changed this a wee bit to better reflect that this is truly an error — we ideally should never see this. And for better debugging, I've added the arguments passed to the method.

annawoodard:
:+1:

PR #351, opened 2016-05-06
Plot resources allocated rather than resources requested.

### LINKED ISSUE #350: Plot resources allocated
We are currently plotting the resources that are initially requested, which is not that illuminating. We should plot the resources allocated by WQ.

annawoodard:
Fixes #350.


matz-e:
Thanks! Is that a new API addition or did I just miss this?

annawoodard:
@matz-e: It's a new addition. Turns out resources requested is just the initial guess we provide.

PR #358, opened 2016-05-10
Allow user to not specify a core guess for WQ allocations.

### LINKED ISSUE #356: Allow WQ to guess cores
We currently are not able to run type 3 configurations in #338, because we set a default number of cores as 1. We rely on that default in `core/source.py` to guess at hunger, etc. We need to be able to leave the cores passed to WQ as `None` while using 1 in  `core/source.py`.

annawoodard:
Default can be `None` now, though we use 1 in that case
for relevant master calculations. Fixes #356.


matz-e:
Thanks!

PR #360, opened 2016-05-11
Specify default category

annawoodard:


lobster/core/workflow.py
```diff
@@ -175,7 +175,7 @@ class Workflow(Configurable):
     def __init__(self,
             label,
             dataset,
-            category=None,
+            category=Category(),
```
matz-e: I didn't provide a default originally since this way, you end up with a category called 'default' for every workflow, not a single category for every workflow.

matz-e: Nevermind, that gets sorted out by the configuration object. I still think it would be better to not set the default name of the `Category` class to 'default' (I think enforcing setting a name is a good practice), and to set the `mode` depending on the name. How about changing this line to `category=Category('default', mode='fixed')`? That way it is immediately clear what is happening and which mode is getting run when the workflow does not have a category specified explicitly. IIRC python would only keep one `Category` object around for this, too… (the way you're never supposed to use `arg=[]` in a constructor)

annawoodard: @matz-e: Thanks! Your suggestion is a big improvement. I'll implement it now.

matz-e: You could also simplify line 198, `self.category = category if category else…`, pretty please?

annawoodard: Good catch, will do.

matz-e:
Thanks!

PR #371, opened 2016-05-20
Cache the dataset info we create from DBS.

### LINKED ISSUE #342: Cache DBS queries
DBS queries could be cached per user, speeding up ntuplizing workflows quite a bit. My current workflow often sits around for ~1 hour just performing DBS queries. If there is a timestamp in DBS we could check, we can cache DBS info in, e.g., `XDG_CACHE_DIR` or so, and re-use it quicker than firing a bunch of queries to the DBS server.

matz-e:
Should make subsequent ntuple runs a lot faster with many datasets.  Watch
out for a new python dependency!
Fixes #342.

### DIFF
--- a/lobster/cmssw/dataset.py
+++ b/lobster/cmssw/dataset.py
@@ -1,12 +1,14 @@
 import logging
 import math
 import os
+import pickle
 import re
 import requests
 from retrying import retry
 import shutil
 import sys
 import tempfile
+import xdg.BaseDirectory
 
 from lobster.core.dataset import DatasetInfo
 from lobster.util import Configurable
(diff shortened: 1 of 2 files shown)


lobster/cmssw/dataset.py
```diff
@@ -37,9 +39,25 @@ def listBlocks(self, *args, **kwargs):
...
-        self.cache = tempfile.mkdtemp()
-    def __del__(self):
-        shutil.rmtree(self.cache)
+        self.cachedir = xdg.BaseDirectory.save_cache_path('lobster')
+
+    def cache(self, name, baseinfo, dataset):
+        logger.debug("writing dataset '{}' to cache".format(name))
+        cache = os.path.join(self.cachedir, name.replace('/', ':')) + '.pkl'
+        with open(cache, 'wb') as fd:
+            pickle.dump((baseinfo, dataset), fd)
```
annawoodard: I can't recall now-- does `listFileSummaries` return a vanilla dictionary? If there are no custom classes, maybe saving this in json format would be preferable, as it is faster and human-readable (I can imagine scenarios where that would be useful.)

matz-e: This saves quite a bit more than just the summaries — basically all the info from subsequent calls to DBS. I had thought about splitting it up into a JSON and a pickle, but opted for simplicity. It would be a bit more complicated to write a custom `DatasetInfo` to JSON and back converter.

annawoodard: Gotcha. Makes sense, just wanted to check. Thanks!

annawoodard:
Looks great. Thanks, @matz-e!

PR #372, opened 2016-05-20
Use more trees to measure the content size of EDM files.

### LINKED ISSUE #361: Baresize should account for run and lumi trees.
I just discovered that one can write data not just to the event tree, but also the run and lumi trees. We should account for that.

FWIW, I started using this feature to only write certain events to the output file, but keep cut histograms and write them at the end of each run (less data than writing one histogram per event).

matz-e:
Uses Events, Lumis, and Runs trees instead of just the Events tree.
Fixes #361.


annawoodard:
Awesome! Thanks, @matz-e!

PR #375, opened 2016-05-27
Write a final newline to report.json.

### LINKED ISSUE #374: Add newline at the end of report.json
Logstash uses newlines to delimit log events, so if an event doesn't end with a newline, Logstash assumes it's not done being written and won't read it. `report.json` ends on a curly bracket instead of a newline, meaning that the final curly brace will never be read and the JSON won't be parsed correctly. This precludes real-time indexing of these files in Elasticsearch.

matz-e:
For ELK. Fixes #374.


annawoodard:
👏 👏 👏

PR #376, opened 2016-05-27
Monitor everything in mainloop

matz-e:
Will hopefully provide a better overview w.r.t. time we spend in either WQ or Lobster.

Broke the fraction plot into two stack plots: one for Lobster, the other for WQ. WQ is setup such that the times should add up to what is called "fetch" for the Lobster plot.

Example:
screen shot 2016-05-27 at 15 02 16

(from [here](http://www.crc.nd.edu/~mwolf3/lobster/ttH/v28/))

### DIFF
--- a/lobster/commands/data/index.html
+++ b/lobster/commands/data/index.html
@@ -148,8 +148,8 @@
             <a href="all/memory-plot.png"><img alt="" src="all/memory-plot.png"/></a>
             <a href="all/disk-plot.png"><img alt="" src="all/disk-plot.png"/></a>
             <h3>Time Breakdown</h3>
-            <a href="all/lobster-fraction-hist.png"><img alt="" src="all/lobster-fraction-hist.png"/></a>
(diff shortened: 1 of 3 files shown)


klannon:
I haven't tested it, but the plot looks very nice.

matz-e:
I'm curious how it'll evolve over time…

klannon:
Indeed.  What I'd really like to see is what happens when Lobster/WQ is dealing with 1-2k (or more) workers.  I think Lobster and WQ previously haven't shown any issues running below that scale, but above 1k workers, it seemed like "other" used to dominate things.

annawoodard:
Awesome! Thanks, @matz-e! I think this will take us far in uncovering the mystery load!

matz-e:
Well… we may just see that our database queries need to be more efficient than they currently are…

annawoodard:
Knowledge is power.

PR #384, opened 2016-06-02
Re-read stored configuration upon changes.

matz-e:
Does away with the current implementation of the configure command, and swaps it out by editing a saved configuration that is synced with the pickled once changes are detected (via the modification time).

Other changes:

* plotting now displays the original configuration
* category runtime is now modifiable
* separate output log for configuration changes
* removed arguments to increase thresholds, as that's better done by changing the configuration directly without a restart

### DIFF
--- a/docs/run.rst
+++ b/docs/run.rst
@@ -135,11 +135,8 @@ the option ``LIBHDFS_OPTS=-Xmx100m``.
 Changing configuration options
 ------------------------------
 
-Lobster provides a ``configure`` command to change certain mutable
-parameters.  When running, the changes may take a while to show any effect.
-Parameters of the :class:`~lobster.core.config.Config` object may be
-changed with simple python expressions.  Example usage::
-
-    lobster configure "storage.output.append(storage.expand_site('root://T3_US_NotreDame/foo/bar'))" config.py
-    lobster configure "advanced_options.threshold_for_failure += 30"
-    lobster configure categories.io_intense.tasks=40
+Lobster provides a ``configure`` command to change certain configuration
+settings.  The command will open an editor showing the current
+configuration.  Changes will be propagated when the configuration is
+re-read by the Lobster main-loop after saving the file.  This may take a
+few minutes for changes to have an effect
--- a/docs/run.rst
+++ b/docs/run.rst
@@ -139,4 +139,6 @@ Lobster provides a ``configure`` command to change certain configuration
 settings.  The command will open an editor showing the current
 configuration.  Changes will be propagated when the configuration is
 re-read by the Lobster main-loop after saving the file.  This may take a
-few minutes for changes to have an effect
+few minutes for changes to have an effect.
+
+Modifiable attributes are listed in the documentation of each class.
--- a/docs/run.rst
+++ b/docs/run.rst
@@ -135,10 +135,21 @@ the option ``LIBHDFS_OPTS=-Xmx100m``.
 Changing configuration options
 ------------------------------
 
-Lobster provides a ``configure`` command to change certain configuration
-settings.  The command will open an editor showing the current
-configuration.  Changes will be propagated when the configuration is
-re-read by the Lobster main-loop after saving the file.  This may take a
-few minutes for changes to have an effect.
+Lobster copies the initial configuration to its working directory as
+`config.py`.  This configuration can be changed to modify the settings of
+a running Lobster instance.  These changes will be propagated when the
+configuration is re-read by the Lobster main-loop after saving the file.
+This may take a few minutes for changes to have an effect, Lobster show
+logging messages about changes in both the main log and `configure.log` in
+the working directory.
 
-Modifiable attributes are listed in the documentation of each class.
+Only attributes mentioned as modifiable in the documentation of each class
+can be changed.
+
+Lobster also provides a ``configure`` convenience command to edit the
+configuration, which will launch an editor to edit the current
+configuration.
+
+.. note::
+   The ``configure`` command uses the environment variable ``EDITOR`` to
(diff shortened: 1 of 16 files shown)


lobster/commands/plot.py
```diff
@@ -1083,9 +1083,7 @@ def make_plots(self, xmin=None, xmax=None, foremen=None):
 
         jsons = self.savejsons(units_processed)
 
-        with open(os.path.join(self.__plotdir, 'config.py'), 'w') as fd:
-            fd.write(str(self.config))
-
+        shutil.copy(os.path.join(self.config.workdir, 'config.py'), os.path.join(self.__plotdir, 'config.py'))
```
annawoodard: :+1: * 100

lobster/ui.py
```diff
@@ -81,4 +81,14 @@ def boil():
         if not getattr(args, "foreground", False):
             logger.removeHandler(console)
 
+    for p in args.plugin.additional_logs():
+        fn = p + '.log'
+        l = logging.getLogger('lobster.' + p)
+        logger.info("saving additional log for {1} to {0}".format(os.path.join(cfg.workdir, fn), p))
+        if not os.path.isdir(cfg.workdir):
+            os.makedirs(cfg.workdir)
+        fileh = logging.handlers.RotatingFileHandler(os.path.join(cfg.workdir, fn), maxBytes=500e6, backupCount=10)
```
annawoodard: I think we should consider reducing the `maxBytes` here (and for all `RotatingFileHandler`s). My current `process.log` is 251 MB with 2.7 million lines and it's starting to get pretty heavy to view.

matz-e: Agreed. Any suggestions? I would go for 100 MB next, but maybe it would be prudent to go for 50-ish? We could increase the `backupCount` to compensate.

annawoodard: I'd vote 100 MB, I don't think we need to go smaller than that.

lobster/util.py
```diff
@@ -127,6 +124,52 @@ def attr(k):
...
+            our_original = self.__kwargs.get(arg, None)
+            theirs = getattr(other, arg)
+
+            if isinstance(ours, Configurable):
+                ours.update(theirs)
+            elif hasattr(ours, '__iter__'):
+                diff = len(ours) - len(theirs)
+                if diff != 0 and not arg in self._mutable:
+                    if our_original != theirs:
+                        logger.error("modified immutable list {}".format(arg))
```
annawoodard: We should document better what is and is not mutable.

lobster/core/config.py
```diff
@@ -170,9 +170,9 @@ class AdvancedOptions(Configurable):
     """
 
     _mutable = {
-            'payload': (None, None, tuple()),
-            'threshold_for_failure': ('source', 'update_paused', tuple()),
-            'threshold_for_skipping': ('source', 'update_paused', tuple())
+            'payload': (None, [], False),
```
annawoodard: It should be noted in the documentation somewhere that these are mutable.

annawoodard: Oops I should have read all the way through the PR first. Nevermind!

docs/run.rst
```diff
@@ -139,4 +139,6 @@ Lobster provides a ``configure`` command to change certain configuration
 settings.  The command will open an editor showing the current
 configuration.  Changes will be propagated when the configuration is
 re-read by the Lobster main-loop after saving the file.  This may take a
-few minutes for changes to have an effect
+few minutes for changes to have an effect.
+
+Modifiable attributes are listed in the documentation of each class.
```
annawoodard: I think we should clarify here that the user has the option to just edit the configuration directly and wait up to 15 minutes for the change to take effect, rather than running the `configure` command. I'm just trying to have sympathy for the new graduate student that doesn't know what the `EDITOR` environment variable is and gets thrown into a VI session and has no idea what the key bindings are and gets stressed.

annawoodard: (see @klannon, we are making sure this is ready to pass on when the time comes)

matz-e: This is (should be) the same editor that `git` will spawn when you don't pass the `-m` flag to the `commit` command… so they'll make contact with it one way or another. FWIW, I've set my `EDITOR` variable to `mg`, a micro-emacs. But we can mention that there's a configuration in the working directory :)

klannon: I think one of the first things every student needs to learn is how to quit out of all the common editors.  Of course, what most students are told is that if you call `commit` and forget `-m` then you're hosed and you need to `^C` out and redo it with the `-m`.  Probably this approach doesn't work here!  So, it would be good to include some tips on specifying an editor that will be more familiar than `vi` for most people.

matz-e: FWIW, I would actually recommend to set the environment variable and **never** use `-m`. By launching and editor, it is actually easier to type in longer commit messages, which in turn may give you better documentation to go with the code (which of course is mostly useful when you also use `git blame`, etc…).

docs/run.rst
```diff
@@ -135,10 +135,21 @@ the option ``LIBHDFS_OPTS=-Xmx100m``.
...
+Only attributes mentioned as modifiable in the documentation of each class
+can be changed.
+
+Lobster also provides a ``configure`` convenience command to edit the
+configuration, which will launch an editor to edit the current
+configuration.
+
+.. note::
+   The ``configure`` command uses the environment variable ``EDITOR`` to
+   determine which editor to use, and uses `vi` as a default.
```
matz-e: @annawoodard Better? I'm hesitant to give a precise time, since, e.g., the runtime can take a lot more to take effect.

annawoodard: Perfect, thanks!

PR #387, opened 2016-06-06
Reorganize task wrapper python script.

### LINKED ISSUE #373: Rework the task wrapper script to quit when one step fails.
I'm currently looking into why my merge tasks fail (it's xrootd access to the storage element, I think).  Currently, out tasks script just proceeds through the steps as if nothing happened. I think we should rework this to stop at the first error. My suggested changes:

1. The main flow of the script should be a bunch of function calls, no real logic (maybe beyond the CMS code distinctions for now), like:

        config = json.load(…)
        data = {…}
        copy_inputs(config, data, …)
        execute_prologue(config, data, …)
        …

2. Instead of using the context to set the error state, wrap the above functions, and update mona-lisa accordingly, if needed:

        @quit_on_failure(exit_code=210)
        def copy_outputs(config, data, …):
           …

matz-e:
Now quits at the first step that fails, while still sending out a dashboard
update, and preparing all needed output files (courtesy of the atexit
module).  All essential functions are decorated to update exit codes and
quit the main script if they fail.

Fixes #373.

**Tested locally, going to start another run to test the last 2 PRs**


matz-e:
Testing so far has been without problems…

lobster/core/data/task.py
```diff
@@ -3,6 +3,7 @@
 from collections import defaultdict, deque, Counter
 from contextlib import contextmanager
```
annawoodard: Looks like we don't need this anymore.

matz-e: I guess while I'm at it, I can run flake8 on it…

matz-e:
Ironed out some more problems. Should be good now, but my run currently is not progressing very fast…

annawoodard:
But you've had some tasks return successfully?

matz-e:
Yes, I did, and the stats also look good so far, so I think this is ready. For failed tasks, it's definitely faster for me finding the point of failure.

annawoodard:
AWESOME. I've looked it over and everything looks good to me. I'm so excited about this improvement! It's going to be a huge help debugging, thanks for doing it! 👏 👏 👏

matz-e:
Glad to hear that :) — I was procrastinating…

PR #388, opened 2016-06-08
Provide two logs: a debug one and an informative one.

### LINKED ISSUE #369: Provide two logs: an informative and a debug one
Currently, our logs are littered with DEBUG statements (at least the way I use Lobster), and it would be nice to have two separate logs: one with entries above DEBUG (for progress information) and one with all entries (for actual debugging).

matz-e:
Adds a separate log for log level DEBUG and up if the log level is set to
DEBUG.  Thus progress is easier to see in the regular log, while
irregularities can still be followed up in the debug log.

Fixes #369.

### DIFF
--- a/lobster/ui.py
+++ b/lobster/ui.py
@@ -2,7 +2,6 @@ from argparse import ArgumentParser
 import logging
 import os
 import sys
-import yaml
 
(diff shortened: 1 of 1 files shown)


annawoodard:
Looks great. Thanks @matz-e! 👏 👏 👏

PR #394, opened 2016-06-13
Factor out and simplify task creation algorithm.

### LINKED ISSUE #283: Should be able to specify a minimum number of queued tasks per category
See title… @klannon is concerned about using every last bit of every worker, so we should be able to always keep a number of "lesser" multi-core tasks queued to shim into unfilled workers.

### LINKED ISSUE #391: Optimize default task creation algorithm.
When creating tasks, we currently ([here](https://github.com/matz-e/lobster/blob/master/lobster/core/config.py#L94) and [here](https://github.com/matz-e/lobster/blob/master/lobster/core/source.py#L365)) cycle through the categories so that the smallest task limit is processed first. If no task limit is specified, the default sort is on the category name. I think we should consider alternate default sorting methods. For example, we could sort by number of units available such that we keep the categories processed as evenly possible. This has two advantages: 1) it is likely that different categories have different bottlenecks, so maximizing the mixing might minimize the per-category bottlenecks, and 2) for chained workflows with different categories at various steps, this would mean that halfway through processing you'd have half of your last step processed, which might be useful if you want to start using output before the entire project is finished.

See, for example, my current situation, where I'm only running digi and reco tasks at a severe inefficiency because I have 513 fit workers but only 251 tasks running, but there are plenty of mAOD tasks that could be run.

This could also be partially addressed by changes on the factory side which I believe @btovar is considering, although I think the points above would still stand.

```
(.lobster) [earth] ~/lobster-spring-16-scale-test/ttW >work_queue_status -A earth.crc.nd.edu 9001
CATEGORY        RUNNING    WAITING  FIT-WORKERS  MAX-CORES MAX-MEMORY   MAX-DISK
digi                247       1002            0         ~4      ~4400     ~22131
reco                  4          0          513         ~5      ~4400      ~4504
```

matz-e:
Basically takes the algorithm that was in place currently, moves it to its own file, and simplifies it quite a lot.  Description of the algorithm in the code comments.

In principal, takes the tasks left per workflow, multiplied by the number of cores for each workflow, and uses that to weigh how many tasks to create for a workflow.

Also add a knob to specify minimum number of queued tasks per category.

Fixes #283, hopefully closes #391.

### DIFF
--- a/lobster/commands/process.py
+++ b/lobster/commands/process.py
@@ -272,7 +272,7 @@ class Process(Command):
                 have = {}
                 for c in categories:
                     cstats = self.queue.stats_category(c)
-                    have[c] = cstats.tasks_running + cstats.tasks_waiting
+                    have[c] = {'running': cstats.tasks_running, 'queued': cstats.tasks_waiting}
 
                 stats = self.queue.stats_hierarchy
                 tasks = task_src.obtain(stats.total_cores, have)
--- a/lobster/core/create.py
+++ b/lobster/core/create.py
@@ -83,6 +83,9 @@ class Algo(object):
         fill_cores = total_cores + max(int(0.1 * total_cores), self.__config.advanced.payload)
         total_workload = sum(workloads.values())
 
+        if total_workload == 0:
+            return []
+
         # contains (workflow label, tasks, taper)
         data = []
         for wflow, (complete, units, tasks) in remaining.items():
@@ -103,6 +106,9 @@ class Algo(object):
 
             needed_category_tasks -= queued[wflow.category.name]['queued']
             needed_workflow_tasks = max(0, int(needed_category_tasks * workflow_fraction))
+
+            if needed_category_tasks <= 0:
+                continue
 
             taper = 1.
             if needed_workflow_tasks < tasks and complete:
--- a/lobster/core/create.py
+++ b/lobster/core/create.py
@@ -75,8 +75,8 @@ class Algo(object):
                 continue
             elif units == 0:
                 continue
-            ncores = wflow.category.cores or 1
-            workloads[wflow.category.name] += ncores * tasks
+            task_cores = wflow.category.cores or 1
+            workloads[wflow.category.name] += task_cores * tasks
 
         # How many cores we need to occupy: have at least 10% of the
         # available cores provisioned with waiting work
@@ -91,11 +91,11 @@ class Algo(object):
         for wflow, (complete, units, tasks) in remaining.items():
             if not complete and tasks < 1. or units == 0:
                 continue
-            ncores = wflow.category.cores or 1
+            task_cores = wflow.category.cores or 1
             category_fraction = workloads[wflow.category.name] / float(total_workload)
-            workflow_fraction = ncores * tasks / float(workloads[wflow.category.name])
+            workflow_fraction = task_cores * tasks / float(workloads[wflow.category.name])
 
-            needed_category_tasks = category_fraction * fill_cores / ncores
+            needed_category_tasks = category_fraction * fill_cores / task_cores
 
             if wflow.category.tasks_max:
                 allowed = wflow.category.tasks_max - sum(queued[wflow.category.name].values())
--- a/lobster/core/create.py
+++ b/lobster/core/create.py
@@ -1,0 +1,123 @@
+from collections import defaultdict
+
+import logging
+
+logger = logging.getLogger('lobster.algo')
+
+
+class Algo(object):
+
+    """A task creation algorithm
+
+    Attempts to be fair when creating tasks by making sure that tasks are
+    created evenly for every category and every workflow in each category
+    based on the remaining work per workflow and cores used.
+
+    Parameters
+    ----------
+        config : Configuration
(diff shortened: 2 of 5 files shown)


matz-e:
FWIW, I started a master with this a while ago, but haven't been able to grab cores so far.

lobster/commands/process.py
```diff
@@ -272,7 +272,7 @@ def sprint(self):
                 have = {}
                 for c in categories:
                     cstats = self.queue.stats_category(c)
-                    have[c] = cstats.tasks_running + cstats.tasks_waiting
```
annawoodard: Yay for using dictionaries which makes things more readable! :+1: * 1000

I'd like to lodge a request that we try to minimize the use of lists (I know that wasn't the case here but it reminds me) when collecting quantities to shuffle around, in favor of named tuples or dictionaries-- something that reminds the reader what the quantity is. That reduces the amount of Agging around/counting we have to do to remember what `info[-13]` is. I don't think the performance hit is worth the loss of readability, but the nice improvements you've made to the master profiling plots (fraction of time spent in various functions) will tell us for sure if we need to make those kind of optimisations.
```
[condorfe] >python -m timeit "data = {'a' : 1, 'b' :2, 'c' : 3}" "for k in 'abc': x = data[k]"
1000000 loops, best of 3: 1.61 usec per loop
[condorfe] >python -m timeit "data = [1, 2, 3]" "for k in data: x = k"
1000000 loops, best of 3: 0.551 usec per loop
```

lobster/core/create.py
```diff
@@ -0,0 +1,129 @@
...
+              workload
+           2. Do the same for the workflow workload versus the category
+              workload
+           3. Use the first fraction to calculate how many tasks should be
+              created for the category
+           4. Adjust for mininum queued and maximum total task requirements
+           5. Subtract already queued tasks
+           6. Calculate how many tasks should be created for the current
+              workflow based on the previously calculated fraction
+           7. Adjust task size taper based on available tasks and needed
```
annawoodard: We should define the various vocabulary words that are used in the logging/plotting page and not documented. For example: `taper`, `units available` vs `units left` (which sound quite similar in meaning), `failed` vs `paused`. Maybe we can add a glossary? Should we open this in a separate issue?

lobster/core/create.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        # Remaining workload
+        workloads = defaultdict(int)
+        for wflow, (complete, units, tasks) in remaining.items():
+            if not complete and tasks < 1.:
+                logger.debug("workflow {} has not enough units available to form new tasks".format(wflow.label))
+                continue
+            elif units == 0:
+                continue
+            ncores = wflow.category.cores or 1
```
annawoodard: So much simpler than my usual `ncores = wflow.category.cores if wflow.category.cores is not None else 1`! I'll pick this up!

I'm usually vaguely aware of this warning from [PEP 8](https://www.python.org/dev/peps/pep-0008/), but I think I take it too far/heed it when it's not necessary:

> Comparisons to singletons like None should always be done with 'is' or 'is not', never the equality operators.

> Also, beware of writing "if x" when you really mean "if x is not None" -- e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!

annawoodard: Now that I compare them, in the longer one it (might?) be slightly clearer what's going on. But I still like yours better :grinning:.

lobster/core/create.py
```diff
@@ -0,0 +1,129 @@
...
+
+        if total_workload == 0:
+            return []
+
+        # contains (workflow label, tasks, taper)
+        data = []
+        for wflow, (complete, units, tasks) in remaining.items():
+            if not complete and tasks < 1. or units == 0:
+                continue
+            ncores = wflow.category.cores or 1
```
annawoodard: Since we have a few `cores`-like variables around, maybe change this to `task_cores` or `cores_per_task` to differentiate it from, for example, `fill_cores`?

annawoodard:
:clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap:
:clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap:
:clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap:
:clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap::clap:

Looks awesome! Just a few minor things for your consideration above and then I think it's ready to go!

lobster/core/create.py
matz-e: Glossary would be good. Maybe we should add a separate page to the docs for the workings/structure of Lobster and add a glossary (maybe sphinx can do the latter automatically)?

lobster/core/create.py
matz-e: Will do.

annawoodard:
Thanks @matz-e!

Issue #397, opened 2016-06-14
Mention that numpy may require manual installation


matz-e:
Since pip seems to get confused by CMS-provided numpy installations.

PR #400, opened 2016-06-16
Add notification about numpy #397

WenzhaoLi:


annawoodard:
Great start!! I just have two minor suggestions:

1) If a single commit fixes an issue, it's best to close the issue via the commit message with one of the keywords (see [here](https://help.github.com/articles/closing-issues-via-commit-messages/)) so that the commit and the issue are clearly linked.
2) Consider elaborating a bit more. Instead of "If Lobster cannot be installed", perhaps you could put "If Lobster installation fails with an error like XXX"

annawoodard:
Final minor suggestion: maybe we should format command-line output differently than regular output? I'll let @matz-e comment.

docs/install.rst
```diff
@@ -89,6 +89,17 @@ easy modification of the source::
     ./install_dependencies.sh
     pip install .
 
+Troubleshooting
+~~~~~~~~~~~~~~~~~~~~~~~
+
```
matz-e: Graphical nitpick: the `~` should only be under the word…

docs/install.rst
```diff
@@ -89,6 +89,17 @@ easy modification of the source::
...
+If Lobster installation fails with an error like 
+
+"Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-G1nd85/matplotlib/",
+
+try reinstalling `numpy`::
```
matz-e: I would also format the error message like code (so that it stands out and looks like it would on the terminal) — maybe also add a few lines of the error message.  Error code 1 can mean all kinds of things, having something mentioning `numpy` in the errormessage is useful. And maybe "manually installing" instead of "reinstalling" would be better? Since it never installed properly in the first place…

matz-e:
I've added some comments (sorry if they are a bit short. Currently dealing with some HCAL stuff before I get back to having a Lobster time-slice). In general, I was thinking this could be a
```
..note ::
  Try manually installing…
```
Right by the lines that deal with the installation. But a "Troubleshooting" section works for me, too.

PR #401, opened 2016-06-16
Update task.py. Removal of spaces from the keys.

WenzhaoLi:


annawoodard:
Let's try this again, as discussed on Slack :)

PR #402, opened 2016-06-16
Update task.py. Removal of spaces from the keys.

WenzhaoLi:
Update task.py. Removal of spaces from the keys.

### DIFF
--- a/lobster/core/data/task.py
+++ b/lobster/core/data/task.py
@@ -972,23 +972,23 @@ def write_zipfiles(data):
 data = {
     'files': {
         'info': {},
-        'output info': {},
+        'output_info': {},
         'skipped': [],
     },
     'cache': {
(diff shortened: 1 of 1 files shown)


annawoodard:
@WenzhaoLi: Have you tested this? It looks like this will break things because you change the first instance of each key but not any subsequent uses of it. You need to search for any uses of the key and change them, too. Also, I don't think we should merge until we address the second part of the issue about filenames as keys when they should be values.

annawoodard:
But, great start! 👏 👏 👏

annawoodard:
Also, see [here](https://github.com/matz-e/lobster/blob/master/Contributing.md) about commit messages.

annawoodard:
@WenzhaoLi I think you have a newer version of this on the way, so I am going to close this PR.

PR #408, opened 2016-06-17
Do not require proxy

annawoodard:


annawoodard:
I've tested this successfully with the following configuration: `/afs/crc.nd.edu/user/a/awoodard/lobster-dev/lobster/examples/simple.py`.

annawoodard:
@matz-e: This isn't how we usually want to do things, but I'm going to go ahead and merge now so that we can get Anna going with local running today. If you have a chance, please do a post-merging review and I will implement your comments on the master.

lobster/core/config.py
```diff
@@ -195,6 +200,7 @@ def __init__(self,
             log_level=2,
             payload=10,
             renew_proxy=True,
+            require_proxy=False,
```
matz-e: I don't think this should be False by default!

annawoodard: @matz-e: Good point. I'll change the default to True. But unless you have objections, I'll leave the handling for the `file` protocol in.

matz-e: Sounds good. Is the `copy_output` bit already in place? I only see changes to the `check_output`? Will be a few more minutes before I have more time

annawoodard: `copy_output` doesn't need to change as long as the `file` protocol is handled by `check_output`.

PR #412, opened 2016-06-20
Move sandbox to a class

matz-e:
Keeps the number of workflow parameters down a bit, and also allows to easily create a basic sandbox for custom usage/non-standard CMSSW usage.

Next step: move more contents of the `cmssw.Sandbox._package` method to the parent.

### DIFF
--- a/lobster/core/workflow.py
+++ b/lobster/core/workflow.py
@@ -139,8 +139,9 @@ class Workflow(Configurable):
             Activates output file merging when set.  Accepts the suffixes
             *k*, *m*, *g* for kilobyte, megabyte, …
         sandbox : Sandbox
-            Path to a sandbox to re-use.  This sandbox will be copied over
-            to the current working directory.
+            The sandbox to use.  Currently can be either a
+            :class:`~lobster.cmssw.Sandbox` or a
+            :class:`~lobster.core.Sandbox`.
         command : str
             Which executable to run (for non-CMSSW workflows)
         extra_inputs : list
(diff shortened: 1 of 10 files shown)


lobster/core/workflow.py
```diff
@@ -139,14 +138,9 @@ class Workflow(Configurable):
         merge_size : str
             Activates output file merging when set.  Accepts the suffixes
             *k*, *m*, *g* for kilobyte, megabyte, …
-        sandbox : str
+        sandbox : Sandbox
             Path to a sandbox to re-use.  This sandbox will be copied over
```
annawoodard: I think this needs to be updated for clarity. You can either re-use a sandbox, or create a new one.

matz-e: Ah, sorry, just changed the type, but not the comment m/

klannon:
By the way, once @annawoodard pointed out my mistake I've been "successfully" using this branch to run my jobs.  I say "successfully" because they all still fail, but not for the problem that trigger this PR.  (Mostly, I've had output and stage out issues; we need to get rid of bestman.)  From my point of view, I think this can be merged with the master.  Any more testing needed first?

matz-e:
Do the latest documentation changes sound acceptable? You could merge it yourself if you agree!

klannon:
Yes, they do.  OK then.  Professors merging pull requests?  What has this world come to?

annawoodard:
@matz-e: Sorry I didn't finish reviewing yesterday. It looks like you missed updating `simple.py`.

matz-e:
Shouldn't need updating — just ran it to pack the sandbox, and it seems to work fine. If no sandbox parameter is passed, nothing should change from the user's perspective.

annawoodard:
Ah, I was just thinking it'd be good to show an example. But I guess that's what readthedocs is for. Got it. Thanks!

matz-e:
Examples should be in `mc_gen.py`

PR #414, opened 2016-06-20
Limit the number of files passed per SQL query.

annawoodard:


matz-e:
Thanks! I recall we have that in place somewhere else already, too!

PR #419, opened 2016-06-21
Don't check filesizes when a list of files is given.

annawoodard:
Closes #287 as 29a3f40 was the only unmerged commit and this re-implements that.


klannon:
Note: Without this, Lobster is pretty much useless for large file lists.  It took me ~8 hours when I was trying to make miniAOD from unmerged AOD output because it had to fire up the hdfs (and therefore Java) 100k times before it could run.  I haven't checked it but I think it should definitely be merged when the checks are satisfied!

PR #420, opened 2016-06-22
Allow variables in paths and set default label.

### LINKED ISSUE #417: Make example configs useable out-of-the-box
screen shot 2016-06-21 at 11 38 56 am

matz-e:
I left the `lobster_` prefix for branding, and since @btovar is actually adjusting the `work_queue_status` display.  The first commit fixes the code, I'll add some more documentation in later commits.

Fixes #417.

### DIFF
--- a/docs/run.rst
+++ b/docs/run.rst
@@ -50,20 +50,15 @@ Workers can be either submitted directly, as shown above, or by using the
 `WorkQueue` factory, which allows for dynamic scaling of the processing.
 The factory uses a config written in JSON, like:
 
-.. code-block:: json
+.. literalinclude:: ../examples/factory.json
+   :language: json
 
-    {
-      "max-workers": 100,
-      "min-workers": 0,
-      "cores": 4,
-      "memory": 3600,
-      "disk": 12000
-    }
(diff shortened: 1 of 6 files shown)


matz-e:
Now here's an additional idea: should we provide a `work_queue_factory` configuration in the `example/` directory?

annawoodard:
Preferably one that matches one of the example configurations.

matz-e:
The cool thing about sphinx is that we can just include/import the one from the examples directory into the docs :)

docs/run.rst
```diff
@@ -53,7 +53,6 @@ The factory uses a config written in JSON, like:
 .. code-block:: json
 
     {
-      "master-name": "lobster_shuffle.*",
       "max-workers": 100,
       "min-workers": 0,
       "cores": 4,
```
annawoodard: Can we change the `memory` requirement to `4400` here to correspond to the `simple.py` configuration, then add the config to the examples and link from here?

After that I vote merge!

matz-e: Yes, we can. I'm going to comment some more on the ELK stuff, then return to this.

annawoodard:
Thanks for doing this!

PR #425, opened 2016-06-24
Update wq members

annawoodard:


lobster/commands/plot.py
```diff
@@ -91,8 +92,10 @@ def mp_call(arg):
     fct, args, kwargs = arg
     try:
         fct(*args, **kwargs)
-    except Exception as e:
-        logger.error('method {0} failed with "{1}", using args {2}, {3}'.format(fct, e, args, kwargs))
+    except:
+        import traceback
+        trace = "".join(traceback.format_exception(*sys.exc_info()))
+        logger.error('{0} failed using args {2}, {3}\n{4}'.format(fct, args, kwargs, trace))
```
matz-e: You can also log the exception with `logger.exception`, which I think will include a traceback. http://stackoverflow.com/a/17397512

docs/install.rst
```diff
@@ -32,7 +32,7 @@ Dependencies
   .. note::
      At Notre Dame, a development version can be accessed via::
 
-      cctools=cctools-109-4883ebaf-cvmfs-40cf5bba
+      cctools=cctools-115-9924592a-cvmfs-40cf5bba
```
matz-e: Don't forget the `tcsh` version at the bottom of the file…

annawoodard: Good catch, thanks!

matz-e:
Awesomer!

lobster/commands/plot.py
annawoodard: Good point about `logger.exception`! In this case though, the handling when you throw multiprocessing in the mix is not straightforward in Python 2.x, see [here](http://bugs.python.org/issue13831).  I'm just reverting the change for now, I'll have another look the next time I need a traceback in this neighborhood 😄

annawoodard:
This should be ready now @matz-e.

matz-e:
Thanks!

Issue #415, opened 2016-06-21
Lobster Wrapper Hard-Codes OSG 3.2 while T3 is at OSG 3.3


klannon:
At the minimum, we should update the hard-coding as OSG 3.2 is deprecated in Aug 2016.  It would be nice not to have the version hard-coded.  For example, Lobster could use the version that's used on the node where the master is run?  Thoughts?

matz-e:
Can we find that out reliably? We could make that an option, and pass it along as an environment variable?

matz-e:
I guess `OSG_LOCATION` could be a good indicator.

klannon:
Right, I think both of those would be desirable.  By default, Lobster could use the value in `OSG_LOCATION` and if that isn't set, then Lobster could issue a message to the user requesting them to set a specific version via a configuration option.  The user could also use the config option if she or he wanted to override the default behavior.  I haven't had a chance to look, but do we already have a method for passing options into the wrapper script?

matz-e:
Yes, we actually do. Should be a quick fix.

klannon:
Well, I don't think it's an emergency so would this be another quick thing for @WenzhaoLi to try?

matz-e:
Sure. We need an option defaulting to `None` [here](https://github.com/matz-e/lobster/blob/master/lobster/core/config.py#L215), some environment checking and member variable setting [here](https://github.com/matz-e/lobster/blob/master/lobster/core/source.py#L145) — throw an `AttributeError` if `OSG_LOCATION` is not in the environment and the configuration variable is not set. Then said member variable needs to be put in the task environment [here](https://github.com/matz-e/lobster/blob/master/lobster/core/source.py#L418) (like `LOBSTER_OSG_LOCATION`), and then, finally, used in the wrapper [here](https://github.com/matz-e/lobster/blob/master/lobster/core/data/wrapper.sh#L128).

Now do we want to be mindful of different RH versions being used? Then there has to be a little bit of shell magic to make everything match up.

klannon:
I think we do want to be a little bit careful for two reasons:
 1.  We can't be sure we'll be running on exactly the same version of Linux as the submitting machine.
 2.  We don't know for sure that the submitting machine will be using OASIS and we don't want to accidentally point our worker job at some locally installed OSG version.

[...]

PR #426, opened 2016-06-27
Support for gsiftp.

### LINKED ISSUE #416: Add gsiftp protocol for stage-out
Until we've scraped every last bit of bestman from the servers, we should provide `gsiftp://` as a more reliable alternative to `srm://`.

matz-e:
Full hookup, including example modifications. Tested with the simple configuration, one failed task, many more successful. Currently lobbed into the `srm` protocol, since the mechanism is nearly the same (exactly the same when using `gfal`, which we should default to when switching to osg 3.3, see #415).

[edited since GH does not support MD in titles]

Fixes #416.


matz-e:
This will make the example configuration a whole lot more robust when not using `chirp`… @klannon would probably agree?

annawoodard:
Awesome! Thanks @matz-e!

annawoodard:
Please resolve conflicts and merge.

PR #434, opened 2016-07-01
Rework initialization

matz-e:
Ensures that less incomplete working directories are created.  Does some minimal filesystem/DBS querying to make sure that dataset names are correct and output directories can either be created or are empty.

### DIFF
--- a/lobster/commands/process.py
+++ b/lobster/commands/process.py
@@ -23,7 +23,9 @@ import work_queue as wq
 
 logger = logging.getLogger('lobster.core')
 
+
 class Terminate(Command):
+
     @property
     def help(self):
         return 'terminate running lobster instance'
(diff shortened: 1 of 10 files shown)


lobster/commands/process.py
```diff
@@ -32,21 +34,25 @@ def setup(self, argparser):
         pass
 
     def run(self, args):
+        self.kill(args.config)
+
+    def kill(self, config):
         logger.info("setting flag to quit at the next checkpoint")
-        logger.debug("Don't be alarmed.  The following stack trace doesn't indicate a crash.  It's just for debugging purposes.")
+        logger.debug("the following stack trace doesn't indicate a crash; it's just for debugging purposes.")
```
annawoodard: 👏 👏 👏

annawoodard:
Awesome! We get things organized better, and for free get one fewer knob (`require_proxy`) for users to need to understand!

PR #435, opened 2016-07-01
ELK Support

yannakopoulos:
Adds optional ELK monitoring support. To enable ELK monitoring, pass an `ElkInterface` object to your `Config` object. Our Elasticsearch cluster is at `elk.crc.nd.edu:9200` and our Kibana instance is at `elk.crc.nd.edu:5601`. Check your log file for a link to your Kibana dashboard.

Issues: 
- If your Elasticsearch config is wrong the first time you start Lobster, Lobster quits. This is intended behavior, but it would be nice if we could fail during config creation so that we notify the user, because right now you'd only know that if you noticed Lobster wasn't running and checked the logs. However, for whatever reason, putting a call to Elasticsearch in `ElkInterface.__init__` breaks `lobster.db`. So check your logs after you start to make sure you didn't fail without noticing. 
- The Elasticsearch check might fail even outside of `ElkInterface.__init__` under certain circumstances (`--foreground`), for some reason.


lobster/commands/elk.py
```diff
@@ -0,0 +1,25 @@
+from lobster.core.command import Command
+
+
+class Elk_Update(Command):
```
annawoodard: Minor, picky item: class names should be CamelCase.

lobster/core/source.py
```diff
@@ -262,7 +265,7 @@ def __find_root(self, label):
 
     def __setup_inputs(self):
         self._inputs = [
-                (self.siteconf, 'siteconf', False),
+                (self.siteconf, 'siteconf', True),
```
annawoodard: Is this change intentional? Maybe it snuck in during a merge conflict or something? We definitely don't want to change `False` to `True` here or we won't be able to update the site configuration.

lobster/monitor/elk.py
```diff
@@ -0,0 +1,536 @@
...
+            User ID to label Elasticsearch indices and Kibana objects.
+        project : str
+            Project ID to label Elasticsearch indices and Kibana objects.
+        populate_template : bool
+            Whether to send documents to indices with the [template] prefix,
+            in addition to sending them to the [user_project] prefix. Defaults
+            to false.
+        modules : list
+            List of modules to include from the Kibana templates. Defaults to
+            including only the core templates.
```
annawoodard: Please document which parameters are modifiable, see [here](https://github.com/matz-e/lobster/blob/master/lobster/core/workflow.py#L27) for an example.

lobster/monitor/elk.py
```diff
@@ -0,0 +1,536 @@
...
+        self.project = project
+        self.populate_template = populate_template
+        self.modules = modules or ['core']
+        self.prefix = '[' + self.user + '_' + self.project + ']'
+        self.start_time = datetime.utcnow()
+        self.end_time = None
+        self.client = es.Elasticsearch([{'host': self.es_host,
+                                         'port': self.es_port}])
+
+        # supposed to check that the Elasticsearch client exists
```
annawoodard: Another very minor thing: not sure if we have a repository policy on this, but I prefer prepending `FIXME` for this type of comment.

lobster/monitor/elk.py
```diff
@@ -0,0 +1,536 @@
...
+            self.end_time = datetime.utcnow()
+
+        self.update_links()
+
+    def check_client(self):
+        try:
+            logger.info("cluster health: " + self.client.cat.health())
+        except es.exceptions.ElasticsearchException as e:
+            logger.error(e)
+            raise AttributeError("could not connect to Elasticsearch cluster" +
```
annawoodard: Another very minor thing. With more than one or two components being added together, I find `str.format()` ([here](https://docs.python.org/2/library/stdtypes.html#str.forma)) a bit more readable. But maybe I'm just an old dog stuck in her ways?

lobster/commands/elk.py
yannakopoulos: I did that at first, but then the command name became `elkupdate` and that seemed less legible to me. I couldn't come up with a better name at the time, though, so I just threw an underscore in.

lobster/core/source.py
yannakopoulos: Definitely unintentional and unnoticed by me. Thanks!

lobster/monitor/elk.py
```diff
@@ -0,0 +1,536 @@
...
+                                             'port': self.es_port}])
+
+    def update_kibana(self):
+        logger.info("generating Kibana objects from templates")
+
+        any_prefix = re.compile('\[.*\]')
+
+        logger.debug("generating index patterns")
+        try:
+            search_index = es_dsl.Search(using=self.client, index='.kibana') \
```
annawoodard: I'm not very familiar with ELK, so this more for my education than feedback for you: why do you have to start with `'[template]'` and later replace it with `self.prefix`? Why can't you start with `self.prefix`?

lobster/monitor/elk.py
```diff
@@ -0,0 +1,536 @@
...
+
+            links_state = json.loads(links_vis['visState'])
+            links_state['params']['markdown'] = links_text
+            links_vis['visState'] = json.dumps(links_state)
+
+            self.client.index(index='.kibana', doc_type='visualization',
+                              id=self.prefix + "-Links", body=links_vis)
+        except es.exceptions.ElasticsearchException as e:
+            logger.error(e)
+    # TODO: change dashboard link time range to end at end time
```
annawoodard: Isn't this done now?

lobster/monitor/elk.py
```diff
@@ -0,0 +1,536 @@
...
+                    self.prefix)
+        try:
+            self.client.indices.delete(index=self.prefix + '_*')
+        except es.exceptions.ElasticsearchException as e:
+            logger.error(e)
+
+    def index_task(self, task):
+        logger.debug("parsing Task object")
+        try:
+            task = dict([(m, o) for (m, o) in inspect.getmembers(task)
```
annawoodard: For readability, since you do this dictionary-ifying a lot, consider factoring it into a function. The function could have an additional argument `skip` or something like that that would take a list of members to skip (the ones you are popping later like `_task`, `command`, etc).

lobster/monitor/elk.py
```diff
@@ -0,0 +1,536 @@
...
+                 'kib_host': self.kib_host,
+                 'kib_port': self.kib_port,
+                 'user': self.user,
+                 'project': self.project,
+                 'populate_template': self.populate_template,
+                 'modules': self.modules,
+                 'prefix': self.prefix,
+                 'start_time': self.start_time,
+                 'end_time': self.end_time}
+        return state
```
matz-e: I hadn't figured this at the time, but something like [the proxy treatment](https://github.com/matz-e/lobster/blob/master/lobster/cmssw/proxy.py#L39) seems a little more clear w.r.t. which part of the state is dropped and explicitly re-added. Plus, it'll make the code scale/behave better in the future :)

lobster/monitor/elk.py
yannakopoulos: That's actually where I'm searching for the template objects (index patterns in this case) that have the ID prefix `[template]`. Once I've found them, I edit them to use indices from the current run and change the ID so they don't overwrite the templates. Most of that is just finding every instance of `[template]` and replacing it with `self.prefix` before reuploading it.

annawoodard: Got it. Thanks!

annawoodard:
👏 👏 👏 

This looks great! I'm super excited to take the new ELK monitoring for a spin. Thanks, @yannakopoulos!

PR #437, opened 2016-07-11
Elk support with locally-stored templates

yannakopoulos:
Instead of storing templates in Elasticsearch, we now store them as part of Lobster. From the user's perspective, the only thing that has changed is the way that templates are created. Now, instead of saving a template to Kibana with `[template]` prefixes, you just make the visualization with your own data. When you're satisfied, you run `./lobster.py elkdownload config.py` to grab the visualizations from Elasticsearch and save them to the templates folder `lobster/monitor/elk_templates`. They can then be committed and pushed to the repo to distribute them to others, who can grab the latest version and run `./lobster.py elkupdate config.py` to see the plots you added.


annawoodard:
You can add the templates to be installed by specifying each file in `package_data` in `setup.py', for example:

```
      package_data={'lobster': [
          'core/data/task.py',
          'core/data/wrapper.sh',
          'core/data/mtab',
          'core/data/siteconf/JobConfig/site-local-config.xml',
          'core/data/siteconf/PhEDEx/storage.xml',
          'core/data/merge_cfg.py',
          'core/data/merge_reports.py',
          'commands/data/index.html',
          'commands/data/gh.png',
          'commands/data/styles.css',
          'commands/data/category.html',
          'monitor/elk_templates/links.json'
      ]},
```

yannakopoulos:
That seems a bit messy, since we could add new template files whenever. I'm looking for a recursive way to add every JSON file in `monitor/elk_templates`, preserving the directory structure.

annawoodard:
I am not aware of a way to add a directory instead of files. You could use `os.walk` to do it dynamically.

yannakopoulos:
Figured it out. This works fine, for the directory structure I have in mind: 

```
        'monitor/elk_templates/links.json',
        'monitor/elk_templates/index/*.json',
        'monitor/elk_templates/**/dash/*.json',
        'monitor/elk_templates/**/vis/*.json'
```

`elk_templates` is generated like this (but with more descriptive filenames):
```
elk_templates/
    links.json
    index/
        index_pattern_template_1.json
        index_pattern_template_2.json
        ...
    core/
        dash/
            dashboard_template.json
        vis/
            visualization_template_1.json
            visualization_template_2.json
            ...
    advanced/
        dash/
            dashboard_template.json
        vis/
            visualization_template_1.json
            visualization_template_2.json
            ...
    other_modules/
        dash/
            dashboard_template.json
        vis/
            visualization_template_1.json
            visualization_template_2.json
            ...
```

annawoodard:
Awesome! Thanks for educating me, I didn't know about wildcards!

annawoodard:
This is very minor, but to have the structure conform to the structure we use elsewhere in the code, consider something like this:
```
monitor/elk
monitor/elk/data
monitor/elk/data/core
monitor/elk/data/advanced
```

lobster/monitor/elk.py
```diff
@@ -135,79 +134,182 @@ def update_client(self):
...
+        try:
+            os.mkdir(self.template_dir)
+        except Exception:
+            pass
+
+        logger.info("getting index patterns")
+
+        try:
+            os.mkdir('{0}/index'.format(self.template_dir))
+        except Exception:
```
annawoodard: I'd recommend being a bit more specific here by having two `except` clauses. If the directory already exists, you'll get an `OSError`. Passing on that is fine. But if it's any other type of error, I assume the function will fail without this directory. So it's better to log the error and exit.

lobster/monitor/elk.py
```diff
@@ -135,79 +134,182 @@ def update_client(self):
...
+            try:
+                os.mkdir(module_dir)
+            except Exception:
+                pass
+
+            logger.info("getting " + module + " visualizations")
+
+            try:
+                os.mkdir('{0}/vis'.format(module_dir))
+            except Exception:
```
annawoodard: same here

yannakopoulos:
Re: directory structure: I'll move the templates to `lobster/monitor/elk/data`. Do you want `elk.py` in `lobster/monitor` or in `lobster/monitor/elk`? The latter makes the import statement `from lobster.monitor.elk.elk import ElkInterface`, which seems kind of ugly.

annawoodard:
You could move `elk.py` to `interface.py`?

annawoodard:
Since we know it's `elk` from the path.

annawoodard:
Something very minor: when building directory paths, consider using `os.path.join` instead of `"xxx".format()`. I know from experience that the latter might be slightly less maintainable down the line if a future dev gets their slashes mixed up or omit a leading slash. Also-- this is subjective-- but in my opinion it's a bit more readable because it conveys clearly that the object is a path.

annawoodard:
It's also a bit easier maintainability-wise because you can add something to the beginning without getting the `{0} {1}` etc out of order or having to re-assign them.

annawoodard:
Also-- this is totally subjective and not very important, so no pressure to change it unless you want to-- but my vote would be for the json filenames to be all lowercase. That would match the style of all our other data files.

annawoodard:
screen shot 2016-07-11 at 12 57 02 pm

Looks like you may want to switch the plot type here?

annawoodard:
screen shot 2016-07-11 at 1 02 10 pm

I see you can resize these by pulling on the bottom left corner. Is there any way to control the default size? If so, consider making these wider so it's easier to read the data and labels.

yannakopoulos:
They're sized as a ratio of the window. On my 1920 x 1080 screen with the window maximized they look fine. What resolution should I be optimizing for? You can also click the little arrow pointing at the labels to hide them.

annawoodard:
screen shot 2016-07-11 at 1 09 16 pm

yannakopoulos:
Yes, I saw that. Haven't fixed it yet, but I think I just accidentally left that one out of the templates. Once I get this directory change done, I'll re-add it.

annawoodard:
> They're sized as a ratio of the window. On my 1920 x 1080 screen with the window maximized they look fine. What resolution should I be optimizing for? You can also click the little arrow pointing at the labels to hide them.

Gotcha. Let's leave things how they are then for now. My home monitor setup is a bit unusual (I only have one monitor at home so I have it vertical and use it for browsing and coding).

annawoodard:
A few more things to consider:

1. Can you not show the legend by default when there is only one data range? For "bytes output", "units left", etc, it's redundant with the y-axis label, and takes up room.
2. Can you show bytes output in a more reader-friendly unit? In the classic plotting, we dynamically choose the unit based on how big the values are-- not sure if that's an option. If not, I'd go with MB.

lobster/monitor/elk/data/core/vis/[template][core]-Task-status-vs.-time.json
```diff
@@ -0,0 +1,10 @@
+{
```
Comment (annawoodard) [inline lobster/monitor/elk/data/core/vis/[template][core]-Task-status-vs.-time.json]: Is there a way to map the value that's being plotted to a label? The numeric value of the task status is not very enlightening. You can find the definitions [here](https://github.com/matz-e/lobster/blob/master/lobster/core/unit.py#L19).

setup.py
```diff
@@ -20,7 +20,11 @@
         'commands/data/index.html',
         'commands/data/gh.png',
         'commands/data/styles.css',
-        'commands/data/category.html'
+        'commands/data/category.html',
+        'monitor/elk/data/links.json',
+        'monitor/elk/data/index/*.json',
+        'monitor/elk/data/*/dash/*.json',
```
annawoodard: Maybe this should be
```
monitor/elk/data/dash/*.json
monitor/elk/data/vis/*.json
```
now?

yannakopoulos: You're correct. I'll fix that now.

annawoodard:
This looks awesome! Thanks @yannakopoulos!!!!
👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏

PR #442, opened 2016-07-23
Do not cache DBS API

annawoodard:


matz-e:
Awesome, thanks for figuring all that out so fast!

PR #448, opened 2016-08-02
Rework filesystem handling

matz-e:
Does a few things:

* Clears filesystem interfaces before daemonizing
* Splits the `StorageElement` class by factoring out a `FileSystem` class — now code is split along functionality, with each class having one use-case
* Renames `default` to `alternative`, since the default filesystem interactions are output SE interactions, not input SE ones.

I don't know anymore what I was thinking implementing the FS code in the first place? I hope this iteration makes it easier to read and a little less magical?

### DIFF
--- a/lobster/commands/process.py
+++ b/lobster/commands/process.py
@@ -6,6 +6,7 @@ import inspect
 import logging
 import logging.handlers
 import os
+import psutil
 import resource
 import signal
 import sys
@@ -147,6 +148,19 @@ class Process(Command):
         signals[signal.SIGTERM] = lambda num, frame: Terminate().run(args)
 
         self.config.storage.deactivate()
+
+        process = psutil.Process()
+        preserved = [f.name for f in args.preserve]
+        openfiles = [f for f in process.open_files() if f.path not in preserved]
+        openconns = process.connections()
+
+        if len(openconns) > 0 or len(openfiles) > 0:
+            logger.error("cannot daemonize due to open files or connections")
+            for f in openfiles:
+                logger.error("open file: {}".format(f.path))
+            for c in openconns:
+                logger.error("open connection: {}".format(c))
+            raise RuntimeError("open files or connections")
 
         with daemon.DaemonContext(
                 detach_process=not args.foreground,
--- a/lobster/commands/process.py
+++ b/lobster/commands/process.py
@@ -137,6 +137,7 @@ class Process(Command):
             ttyfile = open(os.path.join(self.config.workdir, 'process.err'), 'a')
             logger.info("saving stderr and stdout to {0}".format(
                 os.path.join(self.config.workdir, 'process.err')))
+            args.preserve.append(ttyfile)
 
         if self.config.advanced.dump_core:
             logger.info("setting core dump size to unlimited")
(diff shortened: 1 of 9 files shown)


lobster/se.py
```diff
@@ -518,18 +523,28 @@ def _initialize(self, methods):
             else:
                 logger.debug("implementation of master access missing for URL {0}".format(url))
 
+    def deactivate(self):
+        """Clear all storage-element objects.
+
+        Needed when Lobster daemonizes, as filesystem objects may have
```
annawoodard: :+1: For clear explanation to our future selves.

lobster/se.py
```diff
@@ -19,54 +19,30 @@
...
-    """Weird class to handle all needs of storage implementations.
+class FileSystem(object):
 
-    This class can be used for file system operations after at least one of
-    its subclasses has been instantiated.
+    """Singleton class as an interface for filesystem interactions.
 
-    "One size fits nobody." (T. Pratchett)
+    Needs to be configured before first use, with two lists of
+    `StorageElement` implementations.
```
annawoodard: Would be nice to clarify here the purpose of having two lists.

lobster/se.py
```diff
@@ -58,18 +58,18 @@ def switch(*args, **kwargs):
         return switch
 
     @classmethod
-    def configure(cls, defaults, systems):
+    def configure(cls, defaults, alternatives):
         cls._defaults = defaults
-        cls._systems = systems
+        cls._fallback = alternatives
```
annawoodard: Can we consider choosing either `fallback` or `alternative` for the parameter and attribute (how we used to have `systems` and `_systems`? The fewer vocab mappings we have, the easier things are to keep straight.

lobster/se.py
```diff
@@ -58,18 +58,18 @@ def switch(*args, **kwargs):
...
 
     @contextmanager
-    def default(self):
-        tmp = FileSystem._systems
-        FileSystem._systems = FileSystem._defaults
+    def alternative(self):
```
annawoodard: Same here

lobster/commands/process.py
```diff
@@ -148,6 +149,19 @@ def run(self, args):
...
+        openconns = process.connections()
+
+        if len(openconns) > 0 or len(openfiles) > 0:
+            logger.error("cannot daemonize due to open files or connections")
+            for f in openfiles:
+                logger.error("open file: {}".format(f.path))
+            for c in openconns:
+                logger.error("open connection: {}".format(c))
+            raise RuntimeError("open files or connections")
+
```
annawoodard: :+1::+1::+1::+1::+1::+1::+1::+1::+1::+1::+1::+1::+1:

klannon: I think @btovar is suggesting that we sleep before doing this check to keep from flagging as open connections that take a little time to close.  Did I understand that correctly from Slack?

https://ndlobster.slack.com/archives/cctools-debug/p1470247696000020

matz-e: Yes. I'll see about that.

lobster/se.py
```diff
@@ -503,35 +512,41 @@ def _initialize(self, methods):
...
-                Local(path)
+                yield Local(path)
             elif protocol == 'hdfs':
                 try:
-                    Hadoop(path)
+                    yield Hadoop(path)
                 except NameError:
                     raise NotImplementedError("hadoop support is missing on this system")
             elif protocol == 'srm':
                 SRM(url)
```
klannon: Why no `yield` here?

matz-e: Oversight m(

matz-e:
Will post this again a bit reworked…

PR #453, opened 2016-08-10
XRootD StorageElement

klannon:
The motivation behind this branch was to implement the necessary functionality to do file-based processing on a list of files or a directory only accessible via XRootD.  Note, I haven't been able to do any testing of the `remove` or `mkdir` functionality as I need to get access to a server that allows this.  (I'm told this is FNAL, but I need to reset my Kerberos password.)

Also, please look closely at f6d49c7.  It's not strictly related to XRootD, but I couldn't proceed with my testing without fixing what looked to me like a general bug in `TaskProvider`.  Maybe I'm misunderstanding something though.

### DIFF
--- a/lobster/core/data/task.py
+++ b/lobster/core/data/task.py
@@ -389,7 +389,7 @@ def copy_inputs(data, config, env):
                     timeout,
                     "xrdfs",
                     server,
-                    "ls",
+                    "stat",
                     os.path.join(path, file)
                 ]
 
--- a/lobster/core/task.py
+++ b/lobster/core/task.py
@@ -28,9 +28,12 @@ class TaskHandler(object):
         self._units = lumis
         self.outputs = outputs
         self._local = local
-
+        
         self.taskdir = taskdir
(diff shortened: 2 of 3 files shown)


matz-e:
Fixes in f6d49c7 look fine (there's some white-space change that shouldn't be there, but I'm not sure how to stop emacs from doing that).

Otherwise looks good to me, although I'm wondering if having `root://` in the output SE list will cause problems in normal use (not sure if you tested that?)…

If @annawoodard doesn't beat me to it, I'll circle back to this in a little bit.

klannon:
My test had `root:` in the output list after `hdfs` and `file` but before `chirp` and `srm`, and I didn't see any issues with that.

matz-e:
Great, thanks!

matz-e:
Apart from pep-8 inconsistencies, I think this is good to go. We'll fix anything else coming up in the `master`.

Thanks for this!

annawoodard:
Thanks @klannon! 👏 👏 👏 👏 👏 👏

PR #454, opened 2016-08-11
Update lobster version log 

### LINKED ISSUE #433: Log state of Lobster software
It would be good to have a way to somehow log the state of the Lobster software-- preferably the hash of HEAD. It's hard to debug when that info is lost.

WenzhaoLi:
Update the Lobster version representation.
For example: 2.1-a234fea-dirty (the last part marking uncommitted in changes)
Also print the Lobster version in process.log.

close #433

Also replace some special symbols, which cause weird characters for Putty users.
Make it more Windows friendly.

### DIFF
--- a/lobster/util.py
+++ b/lobster/util.py
@@ -278,7 +278,7 @@ def id2dir(id):
     # Currently known limitations on the number of entries in a
     # sub-directory concern ext3, where said limit is 32k.  Use a
     # modus of 10k to split the task numbers.  Famous last words:
-    # "(10k)Â² tasks should be enough for everyone." we use two levels
+    # "(10k)squared tasks should be enough for everyone." ->we use two levels
     # only.
     id = int(id)
     man = str(id % 10000).zfill(4)
--- a/lobster/util.py
+++ b/lobster/util.py
@@ -7,7 +7,6 @@ import os
 import shlex
 import shutil
 import subprocess
-import yaml
 
 from contextlib import contextmanager
 from pkg_resources import get_distribution
@@ -325,6 +324,7 @@ def verify(workdir):
 
 
 def checkpoint(workdir, key):
+    import yaml
     statusfile = os.path.join(workdir, 'status.yaml')
     if os.path.exists(statusfile):
         with open(statusfile, 'rb') as f:
@@ -333,9 +333,11 @@ def checkpoint(workdir, key):
 
 
 def register_checkpoint(workdir, key, value):
+    import yaml
     statusfile = os.path.join(workdir, 'status.yaml')
     with open(statusfile, 'a') as f:
         yaml.dump({key: value}, f, default_flow_style=False)
+
 
 def get_version():
     if 'site-packages' in __file__:
@@ -349,6 +351,7 @@ def get_version():
         os.chdir(start)
         version = '{major}-{head}-{status}'.format(major=1.5, head=head, status=status)
     return version
+
 
 def verify_string(s):
     try:
--- a/lobster/util.py
+++ b/lobster/util.py
@@ -1,5 +1,13 @@
 # vim: set fileencoding=utf-8 :
 
+# Only use default packages here!  This package is included by `setup.py`.
+# Since this happens *before* dependencies are installed, only packages
+# included with the default python distribution should be imported at the
+# top level.
+#
+# If optional packages are needed, they should be included in the function
+# scope.
+
 import collections
 import inspect
 import logging
--- a/lobster/util.py
+++ b/lobster/util.py
@@ -7,6 +7,7 @@ import os
 import shutil
 import subprocess
 import yaml
+import shlex
 
 from contextlib import contextmanager
 from lockfile.pidlockfile import PIDLockFile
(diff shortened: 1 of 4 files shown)


lobster/util.py
```diff
@@ -312,8 +313,15 @@ def verify(workdir):
     if not os.path.exists(workdir):
         return
 
-    my_version = get_distribution('Lobster').version
+    my_version = get_version()
+    #my_version = get_distribution('Lobster').version
```
annawoodard: I'd recommend removing this.

setup.py
```diff
@@ -1,10 +1,16 @@
 #!/usr/bin/env python
 
 from setuptools import setup
+import subprocess
+import shlex
+
+head = subprocess.check_output(shlex.split('git rev-parse --short HEAD')).strip()
```
annawoodard: Why not just re-use the function you added to `util.py`? This is exactly the same.

lobster/util.py
```diff
@@ -312,8 +313,15 @@ def verify(workdir):
...
+    major,  head, status = my_version.split('-')
+    my_version = major
+
     stored_version = checkpoint(workdir, 'version')
+    major,  head, status = stored_version.split('-')
+    stored_version = major
+
```
annawoodard: This is very minor-- I'm only mentioning it FYI. It's conventional in Python to use an underscore to denote a variable that you aren't going to use. That signals to the person trying to understand the code: don't spend any time figuring out what this variable is, it doesn't get used anywhere. In my opinion it makes code a bit more readable, but to each their own. So the lines above could look something like:
```
current_major, _, _ = get_version().split('-')
stored_major, _, _ = checkpoint(workdir, 'version').split('-')
if current_major != stored_major:
    raise ValueError("Lobster {0!r} cannot process a run created with version {1!r}".format(
        my_version, stored_version))
```

annawoodard:
This looks good to me! I just had some small comments. Good job Wenzhao!! 👏 👏 👏 

Now let's wait for feedback from @matz-e. He might have a better idea than changing directories like I suggested, which is kind of kludgy. Note @matz-e that `get_version()` works whether you are running `lobster` after `pip install` or running `python lobster.py`.

lobster/util.py
```diff
@@ -278,7 +278,7 @@ def id2dir(id):
     # Currently known limitations on the number of entries in a
     # sub-directory concern ext3, where said limit is 32k.  Use a
     # modus of 10k to split the task numbers.  Famous last words:
-    # "(10k)Â² tasks should be enough for everyone." we use two levels
+    # "(10k)squared tasks should be enough for everyone." ->we use two levels
```
matz-e: This looks fine in the browser to me — [like here](https://github.com/matz-e/lobster/blob/master/lobster/util.py#L280). If you have to change this, please use `^2`. Further, `putty` seems a very dated way to log in somewhere to me by now. UTF-8 has been a standard/thing for quite a while. Maybe the anniversary update for Windows 10 will make this a little better — I'll give it a try if my machine at home has already updated.

lobster/util.py
```diff
@@ -7,6 +7,7 @@
 import shutil
 import subprocess
 import yaml
+import shlex
```
matz-e: I think that imports should be sorted alphabetically — that way, it is very easy to see if something has been imported already.

lobster/util.py
matz-e: Seconded.

setup.py
matz-e: I concur, again.

lobster/util.py
matz-e: Adding a little: when you comment out code you should always give a reason or explanation. I've wasted long hours of time tripping over something commented out without further information and then going back through source code history trying to figure out *why* this code was changed, and if the old version is somehow important to me.

This is one of the bad practices common in HEP — commented out code should be dead code and removed.

matz-e:
@WenzhaoLi: Thank you! I think this looks good once comments are addressed. I think @annawoodard is spot on with her remarks.

I don't see a simpler version of `get_version` right now. I don't think it matters, either. It's readable, and pretty compact.

setup.py
WenzhaoLi: I thought the `get_distribution()` can only get the message in `setup.py`. So is there any other way to know the installed lobster version? (except `get_distribution()`)

WenzhaoLi:
Thanks! 
Should I undo the commit or make a new commit for the change?

annawoodard:
Make a new commit.

annawoodard:
Then you can push your branch again, it will update the PR.

WenzhaoLi:
I tried importing `util.py` to `setup.py`. But it didn't work. 
```
Processing /afs/crc.nd.edu/user/w/wli12/lobster
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-h_D5Jr-build/setup.py", line 4, in <module>
        from lobster.util import get_version
      File "lobster/__init__.py", line 1, in <module>
        import cmssw
      File "lobster/cmssw/__init__.py", line 1, in <module>
        import dash
      File "lobster/cmssw/dash.py", line 10, in <module>
        from WMCore.Services.SiteDB.SiteDB import SiteDBJSON
      File "/afs/crc.nd.edu/user/w/wli12/.lobster/lib/python2.7/site-packages/WMCore/Services/SiteDB/SiteDB.py", line 8, in <module>
        from WMCore.Services.Service import Service
      File "/afs/crc.nd.edu/user/w/wli12/.lobster/lib/python2.7/site-packages/WMCore/Services/Service.py", line 89, in <module>
        from httplib2 import HttpLib2Error
    ImportError: No module named httplib2
```

matz-e:
You can try to remove the `import cmssw` from `__init__.py` in `lobster`.  If you get errors executing Lobster, try changing import statements from `from lobster import cmssw` to `import lobster.cmssw as cmssw`?

I think it would be good to de-couple this so that we can import the utility functions without additional packages.

PR #457, opened 2016-08-16
Log Process ID in process.log

### LINKED ISSUE #441: Log lobster PID
We should stuff the PID (**after** daemonizing) into the log to allow for quick cross-checks.  The PID is also in the lockfile, but that may not be available after Lobster crashes; So we should log it…

WenzhaoLi:
close #441


matz-e:
Cool, thanks!

PR #464, opened 2016-08-23
OSG update

### LINKED ISSUE #415: Lobster Wrapper Hard-Codes OSG 3.2 while T3 is at OSG 3.3

WenzhaoLi:
Get the OSG version either from the `Advanced Configuration` or the `Environment Variable`. 

Close #415


matz-e:
All looks good and well. Thanks!

PR #468, opened 2016-08-25
Check for illegal project names in ELK constructor

### LINKED ISSUE #467: ELK crashes after startup for illegal project names
It would be nice to check this in the ELK constructor.

annawoodard:
Fixes #467.

### DIFF
--- a/lobster/monitor/elk/interface.py
+++ b/lobster/monitor/elk/interface.py
@@ -126,6 +126,10 @@ class ElkInterface(Configurable):
         self.client = es.Elasticsearch([{'host': self.es_host,
                                          'port': self.es_port}])
 
+        pattern = r'.*[\\/\*\?\"\,\<\>\|].*'
+        if re.match(pattern, project):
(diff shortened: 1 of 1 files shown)


lobster/monitor/elk/interface.py
```diff
@@ -126,6 +126,10 @@ def __init__(self, es_host, es_port, kib_host, kib_port, project,
         self.client = es.Elasticsearch([{'host': self.es_host,
                                          'port': self.es_port}])
 
+        pattern = r'.*[\\/\*\?\"\,\<\>\|].*'
+        if re.match(pattern, project):
```
matz-e: This all makes sense; I'm just wondering if `pattern` without either `.*` and `re.search` instead of `re.match` would make more sense? It would be a bit simpler, AFAIK…

annawoodard: Good point! I'm on it.

matz-e:
Perrrfect! Thanks!

PR #472, opened 2016-08-27
Set minimun task id.

### LINKED ISSUE #455: Set minimum taskid
See cooperative-computing-lab/cctools#1428.

We should use the new functionality to set the minimum ID in case Lobster is terminated and restarted. This will make debugging easier for @btovar because there won't be repeated task ids.

WenzhaoLi:
Fix #455


WenzhaoLi:
@annawoodard Thanks for the detailed elaboration!! I've tested the code. But it'd be great if you can take a look at it.

annawoodard:
@WenzhaoLi Looks awesome! Thanks for putting this together so fast!

PR #483, opened 2016-09-05
Allow disabling of auto output file determination

### LINKED ISSUE #482: Allow user to bypass automatic output determination
In recent commits we added functionality to disable output file return (which we wanted for the proxy/cache server stress test), but that broke the ability to bypass automatic output determination by specifying a list of output file names. I think what we want is:

```
outputs=[] # do not determine outputs, do not copy the output file back
outputs=['output.root'] # do not determine outputs, copy output.root back
outputs=None # determine outputs
```

annawoodard:
Fixes #482.

### DIFF
--- a/lobster/core/workflow.py
+++ b/lobster/core/workflow.py
@@ -366,10 +366,8 @@ class Workflow(Configurable):
         self.version, self.sandbox = self.sandbox.package(basedirs, workdir)
 
         self.copy_inputs(basedirs)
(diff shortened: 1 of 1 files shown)


matz-e:
Forgot to mention… you should document this in the docstring!

annawoodard:
on it

annawoodard:
done

PR #485, opened 2016-09-07
Add a second efficiency plot

### LINKED ISSUE #484: Disentangle inefficiencies in monitoring
In the current Executable CPU / wall time histogram is calculated as
```
                 efficiency = tasks['time_cpu'] / (1. * tasks['cores'] * (tasks['time_processing_end'] - tasks['time_prologue_end']))                                                                                   
```

This makes it difficult to disentangle inefficiencies caused by the task waiting for data to stream from inefficiencies caused by a mismatch between cores the job is asking for and using. We should consider including both this and a version not divided by cores in the monitoring.

annawoodard:
Fixes #484.

### DIFF
--- a/lobster/commands/plot.py
+++ b/lobster/commands/plot.py
@@ -1058,12 +1058,21 @@ class Plotter(object):
                     modes=[Plotter.PROF | Plotter.TIME]
                 )
 
-                efficiency = tasks['time_cpu'] / (1. * tasks['cores'] * (
+                efficiency = tasks['time_cpu'] / (1. * (
                     tasks['time_processing_end'] - tasks['time_prologue_end']))
                 self.plot(
                     [(tasks['time_retrieved'], efficiency)],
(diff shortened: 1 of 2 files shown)


lobster/commands/plot.py
```diff
@@ -1058,7 +1058,7 @@ def make_workflow_plots(self, subdir, edges, good_tasks, failed_tasks, success_t
                     modes=[Plotter.PROF | Plotter.TIME]
                 )
 
-                efficiency = tasks['time_cpu'] / (1. * tasks['cores'] * (
+                efficiency = tasks['time_cpu'] / (1. * (
```
matz-e: Isn't `time_cpu` larger than the pure wall time if we have 100% CPU usage with multi-core?

matz-e: Scratch that — should read the whole diff m(

PR #487, opened 2016-09-10
Provide time breakdown about what happens when tasks are returned.

### LINKED ISSUE #486: Finer granularity in time monitoring
lobster-fraction-stack

We should find out where all that `return` time is going…

matz-e:
Output:
return-fraction-stack

Fixes #486


matz-e:
Empty parts of the plot are due to nothing being returned…

annawoodard:
Awesome, as usual. Thanks for taking care of this! 👏 👏 👏

PR #490, opened 2016-09-12
Adjust and extend efficiency plots.

matz-e:
Instead of running tasks, divide by committed cores for the efficiency
plots.  Also provide a plot which has the CPU time divided by the number of
available cores, to show dips in efficiency due to unused resources.

Maybe similar to #484?

I'll see about adjusting the pie-chart a little, too.

### DIFF
--- a/lobster/commands/data/index.html
+++ b/lobster/commands/data/index.html
@@ -129,6 +129,8 @@
             <h3>Efficiency</h3>
             <a href="all/cpu-wall-hist.png"><img alt="" src="all/cpu-wall-hist.png"/></a>
             <a href="all/cpu-wall-int-hist.png"><img alt="" src="all/cpu-wall-int-hist.png"/></a>
+            <a href="all/cpu-cores-hist.png"><img alt="" src="all/cpu-cores-hist.png"/></a>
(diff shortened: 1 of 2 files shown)


matz-e:
What the pie chart looks like now:

time-pie

(from [one of my latest runs](http://www.crc.nd.edu/~mwolf3/lobster/ttH/v32_scale_v2_all/), which also includes the additional plots)

klannon:
Do you have the "before" plot?

matz-e:
Et voilà:

time-pie-1

Note that the slices may be differently sized, since the number of cores per task is now accounted for.

klannon:
So in the end, the only real change is the idle cores slots.  Everything else looks (at first glance anyway) to scale the same way so the relative contributions look the same.  Am I reading that right?

matz-e:
It is very similar, probably due to a homogeneous task set. While I generated the plots, I had an iteration without the idle core accounting. You can see that the "failed" slice is slightly different, still, which is what I noted at the time.

annawoodard:
Looks good to me. I think it's great that we start digging into this inefficiency caused by idle resources. Thanks @matz-e!!!

PR #492, opened 2016-09-13
Allow relative XrootD inputs

annawoodard:
@klannon please have a look, I think you are the only other person that is using this functionality. If you send me your configuration I can check, but this PR should not break your workflow. It should just enable workflows with relative paths in the `files` being passed to the `Dataset`.

### DIFF
--- a/lobster/se.py
+++ b/lobster/se.py
@@ -410,14 +410,11 @@ class XrootD(StorageElement):
             return False
 
     def ls(self, path):
-        protocol, server, prepath = url_re.match(path).groups()
-        if not prepath.endswith('/'):
-            prepath += '/'
+        protocol, server, _ = url_re.match(path).groups()
         for p in self.execute('ls', path).splitlines():
-            # Fun fact: xrdfs ls returns *absolute* paths but without the protocol.  Strip off the leading part.
-            p = p.replace(prepath, '', 1)
-            logger.debug('from ls(): {}'.format(p))
-            yield os.path.join(path, p)
+            # We need to add the protocol back so that `lfn2pfn` above
+            # can recognize and remove just the pfnprefix.
+            yield "{0}://{1}{2}".format(protocol, server, p)
 
     def mkdir(self, path, mode=None):
         self.execute('mkdir -p', path)


lobster/se.py
```diff
@@ -410,14 +410,11 @@ def isfile(self, path):
...
-            prepath += '/'
+        protocol, server, _ = url_re.match(path).groups()
         for p in self.execute('ls', path).splitlines():
-            # Fun fact: xrdfs ls returns *absolute* paths but without the protocol.  Strip off the leading part.
-            p = p.replace(prepath, '', 1)
-            logger.debug('from ls(): {}'.format(p))
-            yield os.path.join(path, p)
+            # We need to add the protocol back so that `lfn2pfn` above
+            # can recognize and remove just the pfnprefix.
+            yield "{0}://{1}{2}".format(protocol, server, p)
```
matz-e: I'm wondering here… does this yield URLs in the form `root://deepthought.crc.nd.edu//store/blah…`? With the two slashes after the server name? I somehow have it in my head that that is required for xrootd…

klannon: It does!  Without the `//` XRootD interprets the rest of the URL as a "relative" path.  (Which, by the way, is different from the URL standards, I think--fun!)

annawoodard: @matz-e @klannon Yes, as long as the input path in the `StorageConfiguration` has the double slash, it will be preserved by the `xrdfs ls` call. Here's an example from my config:
```
input path: root://deepthought.crc.nd.edu//store/user/awoodard/ttV/2.02/ttZ_gridpacks/
xrdfs call: xrdfs deepthought.crc.nd.edu ls //store/user/awoodard/ttV/2.02/ttZ_gridpacks/
xrdfs returns: //store/user/awoodard/ttV/2.02/ttZ_gridpacks//gridpack.tar_10.xz
XrootD ls yields: root://deepthought.crc.nd.edu//store/user/awoodard/ttV/2.02/ttZ_gridpacks//gridpack.tar_10.xz
```
And then `pfn2lfn` takes care of the rest.

annawoodard:
If we just remove the `prepath` as before then the `ttZ_gridpacks` part of the path is lost.

lobster/se.py
annawoodard: If we just remove the prepath as before then the ttZ_gridpacks part of the path is lost.

PR #495, opened 2016-09-16
Add pythonic hadoop support via snakebite.

### LINKED ISSUE #494: Use snakebite
Moar, better hadoop… https://github.com/spotify/snakebite

matz-e:
Fixes #494.  Kudos to @klannon for digging this up.  This will make
filesystem operation a lot faster by not relying on the `hadoop` command
being available, and not going through the process of starting a JVM for
every filesystem query.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,7 @@ setup(
         'pyxdg',
         'requests',
         'retrying',
+        'snakebite<2.0',
         'WMCore'
     ],
     dependency_links=[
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -45,6 +45,7 @@
         'pyxdg',
         'requests',
         'retrying',
+        'snakebite<2.0',
```
klannon: FYI: In attempt to make sure someone remembers this, we can move to v2.0+ once OSG upgrades to Hadoop v2.2.0, which I think is coming in OSG 3.4.X.  (We're currenly on OSG 3.3.X.)

annawoodard:
This looks great! Thanks @matz-e and @klannon! Hooray for no new-JVMs-per-query!

PR #497, opened 2016-09-18
Don't depend on pycurl directly.

matz-e:
CMSSW ships with it, and that seems more reliable than having the user
install it.


annawoodard:
(but can you fix tests?)

matz-e:
There you go!

annawoodard:
Thanks @matz-e!!

PR #498, opened 2016-09-19
Don't capture tons of stdout that is never used.

matz-e:
Maybe I'm being too nitpicky, but I think that lots of stdout output would be accounted twice when the PSet is misconfigured:

* we save it unconditionally to the process object we return
* we also print it to the "real" stdout, where it gets stored by WQ to be passed back to the master

I think we should only capture the stdout when we need (programmatically speaking), this may reduce the memory usage measured by WQ in extreme cases.

### DIFF
--- a/lobster/core/data/task.py
+++ b/lobster/core/data/task.py
@@ -145,7 +145,7 @@ def run_subprocess(*args, **kwargs):
     logger.info("executing '{}'".format(" ".join(*args)))
 
     retry = kwargs.pop('retry', {})
-    capture = kwargs.pop('capture', True)
+    capture = kwargs.pop('capture', False)
 
     outfd, outfn = tempfile.mkstemp()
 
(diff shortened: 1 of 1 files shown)


annawoodard:
I like this! Minor thing: it looks like we could also disable capturing at lines 406, 417, 453, 481, 595, 618? I know those are small gains but since you've put it in we might as well be consistent and only capture where we need it? (Or maybe reverse the default?)

matz-e:
I was trying to find the items with the biggest stdout — in general, I think those are the main programs executed. I'd favor inverting the default over adding it to another 6 instances of the call.

annawoodard:
@matz-e: I'd vote for inverting the default for consistency, but I'm fine with merging either way.

annawoodard:
Thanks @matz-e! Hooray for less memory usage!

PR #506, opened 2016-10-04
Guard certain dashboard interactions against failures.

### LINKED ISSUE #502: Be resilient against CMSWeb outages
Anna had a recent run crash, because CMSWeb was not reachable (maybe due to an intervention). That shouldn't happen. The user should be notified, but tasks and the master should not fail.

matz-e:
Helps when SiteDB becomes unreachable, but only do so for task registration
and updates.  If a project can't be registered, the dashboard monitoring
should be disabled.

Fixes #502.

### DIFF
--- a/lobster/cmssw/dash.py
+++ b/lobster/cmssw/dash.py
@@ -46,6 +46,20 @@ conf = {
 
 
 def mayfail(errors, retval=None):
+    """Allow decorated methods to fail with certain errors.
+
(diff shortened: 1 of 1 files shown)


matz-e:
I'm disappointed. I can't review my own PR and request changes — but this should be re-done, and provide a default "unkown" username until SiteDB is reachable again.

matz-e:
Since this should be redone, I'm closing it.

PR #511, opened 2016-10-20
Email notification

### LINKED ISSUE #38: Notify user on errors
Often, the lobster master dies quietly without anyone noticing.  We already catch exceptions on a very high level, where we could also notify the user via email or other means.

This may be helpful: https://docs.python.org/2/library/smtplib.html#smtp-example

WenzhaoLi:
Fix #38 

Lobster will send emails to users under these circumstances:
1. Lobster started successfully (Not sure if this is useful. But sometimes Lobster won't start successfully.)
2. Lobster crashed. (Catch most of the crashes)
3. Lobster finished.
The email will also include `workdir` and `plotdir`.
Users need to specify their emails in `Advance Config`. Nothing will happen if no emails are provided.

### DIFF
--- a/lobster/core/config.py
+++ b/lobster/core/config.py
@@ -173,6 +173,8 @@ class AdvancedOptions(Configurable):
             tasks send to it.
         dump_core : bool
             Produce core dumps.  Useful to debug `WorkQueue`.
+        email_address : str
+            The email address you want to receive emails from Lobster.
         full_monitoring : bool
             Produce full monitoring output.  Useful to debug `WorkQueue`.
         log_level : int
@@ -217,6 +219,7 @@ class AdvancedOptions(Configurable):
                  abort_multiplier=4,
                  bad_exit_codes=None,
                  dump_core=False,
+                 email_address=None,
                  full_monitoring=False,
                  log_level=2,
                  osg_version=None,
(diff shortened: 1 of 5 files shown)


lobster/core/config.py
```diff
@@ -173,6 +173,8 @@ class AdvancedOptions(Configurable):
             tasks send to it.
         dump_core : bool
             Produce core dumps.  Useful to debug `WorkQueue`.
+        email_address : str
```
matz-e: I would shorten this to just `email`

lobster/util.py
```diff
@@ -375,6 +377,17 @@ def register_checkpoint(workdir, key, value):
             json.dump(s, f, sort_keys=True, indent=4)
             f.write('\n')
 
+def sendemail(emailmsg, config):
+    you = config.advanced.email_address
+    if you:
+        msg = MIMEText(emailmsg + "\n\n" + "workdir: " + config.workdir + "\n" + "plotdir: " + config.plotdir + "\n\n" + "From Notre Dame Lobster Team")
+        me = 'Lobster@nd.edu'
```
matz-e: What was the conclusion about this in our meeting a few weeks back? I forgot… @klannon do you recall?

lobster/util.py
```diff
@@ -375,6 +377,17 @@ def register_checkpoint(workdir, key, value):
             json.dump(s, f, sort_keys=True, indent=4)
             f.write('\n')
 
+def sendemail(emailmsg, config):
```
matz-e: I was hoping that this could be a separate monitoring module, like `elk`, and live in `monitor/mail`, for example… @annawoodard, @klannon opinions?

matz-e:
Thank you!

PR #513, opened 2016-10-21
Ensure that task timing is always set

### LINKED ISSUE #510: Another Time Accounting Bug...
This seems similar to #280.  I'm seeing times that are effectively +/- infinity (and that add to give ~0):

good-stage-in-hist

good-prologue-hist

This is with `cctools-133-bcc1ac1d-cvmfs-40cf5bba` and `"version": "1.5-9f41635-clean"`.

matz-e:
Most of our timing information is set at the end of a function that is already decorated by `check_execution`.  Move populating the `task_timing` dict there to ensure that any `return` in the middle of wrapped functions still sets the timing information.

Fixes #510.

### DIFF
--- a/lobster/core/data/task.py
+++ b/lobster/core/data/task.py
@@ -236,7 +236,7 @@ def check_execution(exitcode, update=None, timing=None):
                 sys.exit(ecode)
             finally:
                 if timing:
-                    data['task_timing'][update] = int(datetime.now().strftime('%s'))
+                    data['task_timing'][timing] = int(datetime.now().strftime('%s'))
             return result
         return wrapper
     return decorator
--- a/lobster/core/data/task.py
+++ b/lobster/core/data/task.py
@@ -194,7 +194,7 @@ def calculate_alder32(data):
         data['files']['output_info'][fn]['adler32'] = checksum
 
 
-def check_execution(exitcode, update=None):
+def check_execution(exitcode, update=None, timing=None):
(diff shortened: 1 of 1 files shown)


lobster/core/data/task.py
```diff
@@ -220,6 +234,9 @@ def wrapper(data, *args, **kwargs):
                 data.update(update)
                 logger.error("call to '{}' failed, exiting with exit code {}".format(fct.func_name, ecode))
                 sys.exit(ecode)
+            finally:
+                if timing:
+                    data['task_timing'][update] = int(datetime.now().strftime('%s'))
```
klannon: This line is generating the following exception in all tasks I run:
```
  File "task.py", line 1063, in <module>
    copy_inputs(data, config, env)
  File "task.py", line 239, in wrapper
    data['task_timing'][update] = int(datetime.now().strftime('%s'))
TypeError: unhashable type: 'dict'
```
@matz-e: I'll try to figure out what it should be doing, but if you can just spot it, please let me know.

klannon: OK, my first guess is that `update` --> `timing`?  Trying that...

matz-e: Indeed… sorry, I thought I had pushed a fix there.

klannon:
Also ran with this and saw no issues (once the misnomer was fixed.

PR #526, opened 2016-11-16
Add index for task column to unit tables.

matz-e:
This hopefully reduces our task handling time and will make Lobster more responsive.


matz-e:
Any objections to merging this nonetheless? I don't think that this index will hurt us… @WenzhaoLi could you also please vary the parameters a bit to see if we see any performance difference?

annawoodard:
Sounds good to me

PR #527, opened 2016-11-17
Publishing

matz-e:
Was: `# This is not a proper working version!`, but has improved.

To share and re-processes my own generated MC easier, I hacked together a vaguely working version of publishing. This is super-rough, still.

Guidelines:
* [ASO worker](https://github.com/dmwm/AsyncStageout/blob/master/src/python/AsyncStageOut/PublisherWorker.py)
* [DBS utils](https://github.com/dmwm/DBS/tree/master/Client/utils)

Notes:
* Mentioned doing this to Mike
* I found the old structure hard to mentally grasp and tried to streamline the publication process
* A ton of stuff got lost in the process — I tried to litter the code with `TODO`/`FIXME`s
* Migrating parents got thrown under the bus completely for now…
* This is a turd that could use some serious polishing…

### DIFF
--- a/lobster/cmssw/commands/publish.py
+++ b/lobster/cmssw/commands/publish.py
@@ -1,7 +1,6 @@
 # TODO
 # * make sure that primary dataset name is allowed - fail with useful message
 # * use username in PFN construction
-# * allow to specify DBS instance with a command line option
 # * no all CAPS variable names - should be reserved for constants
 # * fix filetype setting
 # * add file, dataset configurations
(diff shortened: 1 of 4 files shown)


lobster/cmssw/commands/publish.py
```diff
@@ -323,165 +139,254 @@ def setup(self, argparser):
...
-                             for workflow in config.get('workflows', [])]
+            args.datasets = [None] * len(args.workflows)
+
+        args.config.storage.activate()
 
-        workdir = config['workdir']
-        user = config.get('publish user', os.environ['USER'])
-        publish_instance = config.get('dbs instance', 'phys03')
-        published = {'dataset': '', 'dbs instance': publish_instance}
+        user = 'matze'  # FIXME get the same way as proxy/SiteDB
```
annawoodard: Until we do this the same way as proxy/SiteDB, could we read it in from the config rather than hard coding it?

annawoodard:
> I found the old structure hard to mentally grasp and tried to streamline the publication process

I'm so sorry and embarrassed about this code. Thank you so much for doing this!

I think this looks good and since it's already a gargantuan improvement and we're working with limited time we should go ahead and merge it...

matz-e:
> I'm so sorry and embarrassed about this code. Thank you so much for doing this!

It was very thorough, no need to be embarrassed… I just had a hard time getting the bits and pieces together in my mind. BTW, I constantly re-write my own code. I often get better design schemes after a year or so of burying my own code.

lobster/cmssw/commands/publish.py
matz-e: I said this was not really working ;) We could do that/make it a command line argument.

matz-e:
I've published a few datasets with this now, and while it's not perfect, I think it's a good working approximation so far. With the addition of getting the username from SiteDB or a command line argument, I think it's worth removing the "do-not-merge" label.

matz-e:
So… in theory that last commit would enable us to run CRAB jobs over datasets that we publish. *With the one exception that we're blacklisted in the CRAB server.* Whitelisting does not help, but at least all the metadata is found properly…

FWIW, this allows us to treat our MC production just like any other dataset, which I personally prefer. I've ntuplized and re-aod'ed this way, and I find it working very smooth!

Any more comments on this PR? @klannon, @annawoodard?

annawoodard:
@matz-e: it will be a while until I have time to do a detailed code review of this PR, so I think we better just go ahead and merge it. Looking over it quickly, everything looks great. Thank you so much for putting this together.

annawoodard:
@matz-e: Do we know if there's a way to get un-blacklisted? It would be great if collaborators could run over our datasets with CRAB...

matz-e:
I'm trying some CRAB settings. I'll update when the jobs actually run… they're still queued.

matz-e:
You can try
```python
config.Site.ignoreGlobalBlacklist = True
```
But my job is still queued, so I don't think this really works. As an alternative, with `ignoreLocality`, you can run over data as long as you copied it to an outside T2/T3.

PR #528, opened 2016-11-21
Better sqlite handling

### LINKED ISSUE #319: Find lost events

matz-e:
Supersedes #320 and #526.

Fixes #319, closes #320, closes #526.

# Content

I went through `unit.py` and looked at all the where clauses to find the following worthwhile to add in terms of indices:

* units should be indexed on (file, status)
* tasks should be indexed on (workflow, status)
* workflows should be indexed on (label,)

In addition, I think that @annawoodard's work from earlier this year greatly simplifies and lessens SQLite queries while creating tasks, so I rebased it on top of the indices.

# Before changes

tasks running
time split

# After changes

tasks
time split

# My observations

While these two sets of plots are not directly comparable since the "before" has a higher load, it is notable that *what* we spend our time on for a comparable load has shifted away from the database handling to dealing with the processing of tasks of itself.

@WenzhaoLi more testing would be greatly appreciated. Also some verification that with the simplified database handling nothing changed…

### DIFF
--- a/lobster/cmssw/dataset.py
+++ b/lobster/cmssw/dataset.py
@@ -205,6 +205,6 @@ class Dataset(Configurable):
             logger.debug("split lumis detected in {} - "
                          "{} unique (run, lumi) but "
                          "{} unique (run, lumi, file) - "
-                         "enforcing a limit of one file per task".format(dataset, total_lumis, result.total_units))
+                         "enforcing a limit of one file per task".format(self.dataset, total_lumis, result.total_units))
 
         return result
--- a/lobster/cmssw/dataset.py
+++ b/lobster/cmssw/dataset.py
@@ -159,6 +159,7 @@ class Dataset(Configurable):
         result = self.__cache.cached(dataset, baseinfo)
         if result:
             return result
+        total_lumis = sum([info['num_lumi'] for info in baseinfo])
 
         result = DatasetInfo()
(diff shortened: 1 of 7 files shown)


matz-e:
Any comments? I've been running with this for the last few days now, and it runs pretty smooth. I just enabled merging with the two last projects, and any spikes I see in the return fraction seem to be strongly correlated with cleanup, I assume due to successful merge tasks.

WenzhaoLi:
Since there is no complaint, I guess this PR is good to go?

annawoodard:
Yes, it looks great to me. Thanks for doing this and for the careful explanation @matz-e!!!

matz-e:
Well, thank you for laying the ground-work, which was most of the work overall!

PR #535, opened 2016-12-06
Subtract masked lumis when calculating left ones.

matz-e:
Enables us to run on all lumis when using a mask, instead of stopping to create work.


matz-e:
I think this is ready now…

PR #537, opened 2016-12-13
Fix range error

annawoodard:


lobster/core/data/task.py
```diff
@@ -587,37 +587,38 @@ def copy_outputs(data, config, env):
...
-                if p.returncode == 0 and check_output(config, localname, remotename):
-                    transferred.append(localname)
-                    match = server_re.match(args[-1])
-                    if match:
-                        target_se.append(match.group(1))
-                    data['transfers']['srm']['stageout success'] += 1
-                    break
-                else:
-                    data['transfers']['srm']['stageout failure'] += 1
+                if len(prg) > 0:
```
matz-e: Instead of an `if` statement and lots of indented lines, I think it would be better to have an
```
if len(…):
    …
elif len(…):
    …
else:
    # some failure mode

…code as before…
```
here. That would keep indentation lower and less levels of complexity to track when reading the code?

annawoodard: Oops, I meant for 0a21810 to be two separate commits, but I'll just leave it to not make things more confusing. Let me know if it doesn't address your comment satisfactorily.

matz-e:
I've pushed the one thing that I think should be clearer, this is good to merge if you're OK with it.

annawoodard:
Ah, great point @matz-e! I didn't misunderstood your suggestion, sorry about that.

PR #539, opened 2016-12-14
Fail when sourcing the environment fails.

matz-e:
Otherwise, these tasks fail down the line, e.g., with stage-in/out errors.


annawoodard:
Looks good to me, thanks @matz-e!

Issue #521, opened 2016-11-11
Have Lobster e-mail users before proxy expires


WenzhaoLi:
Have Lobster e-mail users 24-48 hours before proxy expires. 
Maybe send e-mail every 12 hours staring 48 hours out.

PR #541, opened 2016-12-20
Send email to user when proxy is going to expire within 24 hours.

WenzhaoLi:
Send email to user when proxy is going to expire within 24 hours.
The email will only be sent once. But every time user renews the proxy, the email will be reset. Which means, if the proxy drops to 24 hours again, another email will be sent, and so forth.

### DIFF
--- a/lobster/cmssw/proxy.py
+++ b/lobster/cmssw/proxy.py
@@ -60,3 +60,6 @@ class Proxy(Configurable):
 
     def expires(self):
         return int(time.time()) + self.__proxy.getTimeLeft()
+
+    def time_left(self):
+        return self.__proxy.getTimeLeft()
--- a/lobster/commands/process.py
+++ b/lobster/commands/process.py
@@ -277,7 +277,9 @@ class Process(Command, util.Timing):
                 if self.config.advanced.proxy:
                     expiry = self.config.advanced.proxy.expires()
                     proxy_time_left = self.config.advanced.proxy.time_left()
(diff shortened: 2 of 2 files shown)


WenzhaoLi:
These commits have been tested.

WenzhaoLi:
close #521

matz-e:
You should fix there style issues (from the tests):
```
./lobster/commands/process.py:280:53: W291 trailing whitespace
./lobster/commands/process.py:282:77: W291 trailing whitespace
./lobster/commands/process.py:283:122: E251 unexpected spaces around keyword / parameter equals
./lobster/commands/process.py:283:124: E251 unexpected spaces around keyword / parameter equals
```

Something like [flycheck](http://www.flycheck.org/en/latest/) may help you. Seems more of a pain with emacs than with vim…

WenzhaoLi:
Thanks for the info. But I am using Putty.. I will find something compatible with Putty. So how did you get the style debug? Can I get it on Github?

matz-e:
It may work on the command-line, too. I have an equivalent in vim in the terminal. If you click on "Details" by the failed checks, you should be able to find the failed test.

WenzhaoLi:
Got it! Thanks!!

klannon:
I've been running with this branch merged into master for a couple of days now.  No problems and I got my proxy e-mail reminder right on schedule.  I'm going to go ahead and merge this.

WenzhaoLi:
Yes, I set a variable to reset the email. I am glad that it's working. Thanks!

PR #562, opened 2017-02-22
Workflow summary merge tweaks

annawoodard:
Before:
before
@muell149's config, merging enabled for all workflows

After:
after
 @muell149's config, merging enabled for all workflows

overview
test run, one workflow with merging enabled

category
test run, one workflow with merging enabled

This tweaks the workflow summary so that files which are already at the target size are included in the count of merged files.


matz-e:
Great, thanks!

PR #564, opened 2017-02-23
Clarify paused failed skipped

annawoodard:
In this PR, I tweak the name/meaning of `paused`, because it doesn't give a hint that it is related to parents. I think our nomenclature is fine internally, but for exposing to the users, I make the proposal below so the three groups (`paused`, `failed`, `skipped`) are orthogonal. It is a bit non-optimal because we can't distinguish `missing_upstream=10` due to one missing unit in each of 10 upstream workflows (causing 1 missing unit of the workflow in question), or 10 missing units in one upstream workflow (causing 10 missing units of the workflow in question). We could specify both how many units are missing in upstream workflows (giving the user a sense for how much work would be required to get to 100%), and how many units in the current workflow that maps onto (giving the user a sense for how big an impact the missing work has). But I think adding a column would just require more effort on the user's part to understand the table, so it makes sense to omit it for now.

before:

| label       | definition |
| -------- | --------- |
| failed | units of this workflow which failed more than threshold for failure |
| skipped | units of this workflow who's input file was skipped more than threshold for skipping |
| paused | failed + skipped plus units which were failed or skipped in all upstream workflows|

after:

| label       | definition |
| -------- |---------- |
| failed | units of this workflow which failed more than threshold for failure |
| skipped | units of this workflow who's input file was skipped more than threshold for skipping |
| missing upstream | units which were failed or skipped in all upstream workflows|

Note 1: I'm proposing `missing upstream` rather than `missing parent` (which would be preferable because it's in line with our configuration nomenclature) because it implies that it includes multiple levels upstream, not just the immediate parents (which is what I think of intuitively when I hear parent).

Note 2: I've also tweaked the other headers to make it clear what quantities refer to events and which to units. This makes the table a bit busier, maybe there's a clearer alternative? I'd argue we should somehow indicate that information though, just hinting by having some words capitalized and not others is not straightforward.

Demo [here](http://www.crc.nd.edu/~awoodard/lobster/test_20170223_1547/).

tooltip


annawoodard:
Actually, maybe we should change `units_paused` to `units_stuck` everywhere. To me paused sounds like they are in that state intentionally, waiting for us to press the play button. I think the connotation of `stuck` matches this case. Then also the internal nomenclature would match the workflow summary better.

matz-e:
I prefer the less busy
```
Events 	read 	written 	Units 	unmasked 	written 	merged 	paused 	failed 	skipped 	Progress 	Merged
```
We could change it to (and at least keep the capitalization)
```
Events 	read 	written 	Units 	unmasked 	written 	merged 	stuck 	failed 	skipped 	Progress 	Merged
```
We could have a two column header, I think that may be clearest?  I'll see if I can come up with something there… like this:
```
                Events                  |                                           Units                                                                     | Progress 	Merged

total 	read 	written |	total 	unmasked 	written 	merged 	stuck 	failed 	skipped 	```
That would be my biggest objection. Also is there a way to get the tooltip to not be in bold?

The point of having paused include all units that were missing to completion was that you should be able to to `unmasked - (written + paused) = units left`. With your change, we'd now have to sum up more numbers… maybe that's fine?

Currently, all upstream numbers are not included because (a) that's hard and (b) for this workflow, that number of units are missing due to units stuck in the parent workflow. As in, stuck plus the rest of the count should give you total, and by including all stuck upstream units, we would exceed that. It would also break the totals column.

**Comment by annawoodard** [comment]
> We could change it to (and at least keep the capitalization)
```
Events 	read 	written 	Units 	unmasked 	written 	merged 	stuck 	failed 	skipped 	Progress 	Merged
```

I feel strongly that expecting the user to understand the meaning of the header based on the pattern of capitalization is asking too much and we should expect normal users to pick up on that subtlety. I'll implement your two-column header and see what you think.

> The point of having paused include all units that were missing to completion was that you should be able to to `unmasked - (written + paused) = units left`. With your change, we'd now have to sum up more numbers… maybe that's fine?

Units left is a very important quantity, do you have an objection to me just adding a column for it in the table?

Ahh, good point about all the upstream numbers, I had mis-remembered that, I should have checked the code, sorry. I think the referring only to the units of the current workflow that are skipped is more straightforward anyways. I'll correct the description.

**Comment by annawoodard** [comment]
@matz-e I've made a first pass at addressing your concerns. Main changes: I've removed the bold tooltips, re-arranged the headers as you suggested, and corrected the stuck units description. You mentioned that it might be annoying for users to sum up the units left, so I've added a column in the summary which corresponds to the `units left` that's reported in the debug log.

Demo updated [here](http://www.crc.nd.edu/~awoodard/lobster/test_20170223_1547/).

**Comment by matz-e** [comment]
Looks good. But the actual header for units left is missing :)

I agree that the column headers as they are now are not optimal. I would hope that with two rows we can make it less ambiguous. I can help out, too, since I'm the nitpicker here.

**Comment by matz-e** [comment]
Maybe your regular plotting just overwrote the custom run you did?

**Comment by annawoodard** [comment]
Oops, indeed, I mixed up branches there :smile:.

Here's one I won't write over: [http://www.crc.nd.edu/~awoodard/lobster/clarify-paused-failed-skipped-demo/](http://www.crc.nd.edu/~awoodard/lobster/clarify-paused-failed-skipped-demo/).

File: lobster/commands/data/styles.css [hunk 1]
```diff
@@ -77,6 +78,12 @@ nav li.inactive:hover {
 
 .lobster {font-family: 'Lobster Two', cursive;}
 
+.code {
+    font-family: 'Share Tech Mono', monospace;
```
matz-e: I'd like to add Fira Code here (based on Mozilla work) here…

lobster/commands/data/styles.css
```diff
@@ -243,3 +259,26 @@ ul.img-list li:hover span.text-content {
...
+.tooltip {
+    position: relative;
+    display: inline-block;
+    border-bottom: 1px dotted black;
+}
+
+.tooltip .tooltiptext {
+    visibility: hidden;
+    width: 340px;
+    background-color: black;
```
matz-e: I've got to note that I find white text on black very hard on the eyes. That's why the headers themselves are kind of muted. Maybe we could switch that to a wee bit lower contrast?

lobster/commands/data/styles.css
```diff
@@ -243,3 +259,26 @@ ul.img-list li:hover span.text-content {
...
+    color: #fff;
+    text-align: center;
+    padding: 5px 0;
```
matz-e: Could we increase the padding? Like 10px all around?

annawoodard:
@matz-e: thanks for the feedback! Demo is updated with tweaks implemented.

matz-e:
Awesome, many thanks!

PR #565, opened 2017-02-24
Do not give up when snakebite encounters an error

annawoodard:
This will be slower, I guess. Maybe there's a way to optimize it? I think it should work for now though.

@klannon @matz-e


matz-e:
Basically reverting a02ff0d? I think we should investigate why this does not work when cleaning up during the run, but succeeds when validating.

annawoodard:
Good point! Sorry, I did not look closely at the history of that bit of code. I'll look at different approaches.

PR #566, opened 2017-02-25
Delete files, even if some do not exist

### LINKED ISSUE #558: Files not cleaned up after merging
Reported by @muell149 on 02/22/2017.  Symptom is that some files that have been merged are not cleaned up.  Not sure why?  The log file shows the call to remove the files after merging.  Snakebite problem?

annawoodard:
This commit includes two major changes:

1) If the snakebite `remove` call fails because a file is missing, try
   again, one at a time, so that files that do exist are removed.
2) Breaks the `fs.remove` call in two: once for failed tasks, which are
   likely not to exist, and once for merged/input cleanup tasks, which
   are likely to be where we expect them. In that way we usually only
   end up in the `except` clause for failed tasks, which should mitigate
   the slowdown from 1).

Fixes #558.

### DIFF
--- a/lobster/core/source.py
+++ b/lobster/core/source.py
@@ -437,7 +437,9 @@ class TaskProvider(util.Timing):
         return tasks
 
     def release(self, tasks):
-        cleanup = []
+        fail_cleanup = []
+        merge_cleanup = []
+        input_cleanup = []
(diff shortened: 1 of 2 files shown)


matz-e:
Awesome!

PR #576, opened 2017-03-23
Add a hack example

annawoodard:
As requested by @klannon.

This explanation isn't very useful. I just wanted to get something in there as requested before I forget and flesh things out more as the section grows.

### DIFF
--- a/docs/trouble.rst
+++ b/docs/trouble.rst
@@ -56,11 +56,24 @@ Changing an immutable attribute in the pickled configuration
 
 You should never need to do this. You should only change mutable configuration
 attributes using the ``configure`` command via `lobster configure config.py`.
-If you're set on doing this, first ask yourself why before how. But if you're
-desperate, here's an example which changes the `label` attribute::
+The point of the `PartiallyMutable` metaclass is to restrict configuration
+options from changing unless they have been declared mutable and a callback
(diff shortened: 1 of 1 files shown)


docs/trouble.rst
```diff
@@ -44,3 +44,25 @@ Mixing DBS and local datasets
...
+
+You should never need to do this. You should only change mutable configuration
+attributes using the ``configure`` command via `lobster configure config.py`.
+If you're set on doing this, first ask yourself why before how. But if you're
+desperate, here's an example which changes the `label` attribute::
+
+    from lobster.core import config
+    from lobster import util
+
+    cfg = config.Config.load("/path/to/working/directory")
```
matz-e: I had a script that I deleted… what I did right there in addition:
```python
import datetime
import os
import shutil

wdir = "/path/to/working/directory"
shutil.copy(os.path.join(wdir, "config.pkl"), os.path.join(wdir, "config.pkl.{:%Y-%m-%d_%H%M%S}".format(datetime.datetime.now())))
```
Never trust a manual backup requirement. Believe me. You will fail, bigly (as have I). Best to have too many backups, as this ruin the run if something goes wrong.

annawoodard: Great point, @matz-e! Thanks for the feedback. Should be addressed now.

PR #578, opened 2017-04-14
Add an explanation if `UnitStore` can't find DB

annawoodard:
I'm not sure if this explanation is too long, or clear enough. Feedback appreciated.


WenzhaoLi:
It looks good to me!

matz-e:
Still pretty verbose, but oh well! Kudos for doing everything via the `argparse` mechanism!

matz-e:
And thank you for making us more user-friendly! We tend to fail in abstruse ways that we shouldn't really… 👏 🎉

Issue #207, opened 2015-09-25
Custom commands do not get input files passed


matz-e:
We should append the list of input files to the custom command, otherwise this seems somewhat pointless?  Currently, this is only done for merge jobs.  Thoughts/comments, @annawoodard?

annawoodard:
Yes, I agree! I've been parsing it out of parameters.json, which is clearly not the user friendly solution :).

matz-e:
As far as I remember, we had a private conversation about this, where we found that we could have the following specification:
```
task:
  - label: foo
    cmd: "cat {files}|grep some_random_string>output"
```

And I think @annawoodard also proposed that we could apply this to `cmRun`:
```
task:
  - label: MyMessyTTH
    cmd: "cmsRun ntuplize.py"
    dataset: "/ttH*/*/MINIAOD"
```
Here we could condense three parameters into one: `cmd`, `pset`, and `parameters`.  For CMS jobs, we can split the command via `shlex` IIRC, and then remove a maybe already present `-j`, add our own for the job report, and extract the parameter set name, and add it to the inputs. Input files can then be used in the command, as long as the worker does:
```
cmd.format(files=map(pipes.quote, input_files))
```
and we execute `cmd` as a shell string.

PR #585, opened 2017-05-23
Allow custom merge commands

### LINKED ISSUE #580: Allow custom merge commands
We allow non-root output files, so it would be convenient (and straightforward) to allow users to specify a custom merge command in that case (for which cmssw and hadd won't suffice).

annawoodard:
Fixes #580.

### DIFF
--- a/lobster/core/workflow.py
+++ b/lobster/core/workflow.py
@@ -177,11 +177,12 @@ class Workflow(Configurable):
         globaltag : str
             Which GlobalTag this workflow uses.  Needed for publication of
             CMSSW workflows, and can be automatically determined for these.
-        edm_output : bool
-            Autodetermined when outputs are determined automatically.
-            Tells Lobster if the output of this workflow is in EDM format.
-            If `True`, cmssw will be used to merge output files. Otherwise,
-            `hadd` will be used.
+        merge_command : str
+            Accepts `cmsRun`, `hadd`, or a custom command. Tells Lobster what
+            command to use for merging. If outputs are autodetermined, cmssw
+            will be used for EDM output and hadd will be used otherwise. Custom
+            commands should accept the output file as the first argument followed
+            by one or more input files to merge.
     """
     _mutable = {}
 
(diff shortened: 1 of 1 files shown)


lobster/core/workflow.py
```diff
@@ -433,12 +434,16 @@ def adjust(self, params, env, taskdir, inputs, outputs, merge, reports=None, uni
...
                 pset = os.path.join(os.path.dirname(__file__), 'data', 'merge_cfg.py')
             else:
-                cmd = 'hadd'
-                args = ['-n', '0', '-f', self.outputs[0]]
+                cmd = self.merge_command
+                args = [self.outputs[0]]
+                if self.merge_command == 'hadd':
+                    args = ['-n', '0', '-f'] + args
+                else:
+                    inputs.extend((i, os.path.basename(i), True) for i in self.extra_inputs)
```
matz-e: Should custom merge commands get the extra inputs? I'd think that that would not be desired in all cases?

matz-e:
BTW, thank you for tackling this! I had been thinking about the need of custom merges before, but the need never arose for me, and I'd have over-engineered it…

lobster/core/workflow.py
annawoodard: Sorry about the delay @matz-e. The motivation for including the extra inputs was to give the user the flexibility and have both of these cases handled:
1) The user wants to use a custom script that isn't in their sandbox 
2) The user wants to use an executable from their sandbox.

We have a few options:
a) Current approach (let user include it in `extra_inputs` and take an efficiency hit transferring unneeded files if `extra_inputs` is specified for processing jobs)
b) Don't support case 1
c) Add an additional, optional parameter to explicitly handle case 1)
d) Try to infer what to do from the `merge_command` the user passes.

My slight preference is for a) because I prefer to keep things simple: c) is extra cognitive load on the user to figure out the meaning of an additional parameter, and d) seems like it could introduce inflexibilities and cause behavior that would be difficult for users to understand without looking at the source.

annawoodard: My close runner-up is c), we could add `merge_extra_inputs` and document they would only be transferred for merge jobs.

matz-e: Thanks for expanding! One more thing, which can be addressed after merging: Since the custom merge command may be something like `tee -a <output> <input1> <input2> …` and thus `merge_command='tee -a'`, we may have to call `shlex.split()` on the command?

IIRC, this would also feed into a discussion we had a while back about running custom commands in the processing workflow?  See #207, which should also take merge commands into account, so that the merge command cut be something like `cat <input1> <input2> … > <output>`.

matz-e:
Thanks! 👏 🎉

PR #599, opened 2017-08-03
Allow WQ port customization.

khurtado:
This is useful if you need your master to be on a specific port number (e.g: an open TCP port, so workers from outside your local network can connect without ssh-tunnels or other tricks).

I wasn't sure where to put that, so I added it as an advanced parameter and kept the "-1" as default.


matz-e:
Thank you 😄

PR #601, opened 2017-09-13
Add "IndexError" to the list of errors in the siteconf frontier try-except clause

khurtado:
Sometimes, SITECONF might exist without a frontier proxy defined, so `siteconf.frontierProxies[0]` will fail with "IndexError: Out of range".


matz-e:
Thank you! 🎉

PR #604, opened 2017-09-28
Do not import snakebite if it is not needed

### LINKED ISSUE #602: Catch snakebite exceptions on CI
See title… current builds fail due to snakebite.

annawoodard:
This is kind of a hack to fix #602, but it seems reasonable not to spend too much time on it since we're not testing snakebite with Travis anyways. It could be a symptom of a bigger dependency issue, but let's solve that problem if it actually comes up, and have working CI in the meantime.

### DIFF
--- a/lobster/se.py
+++ b/lobster/se.py
@@ -2,8 +2,9 @@ import logging
 import os
 import random
 import re
-import snakebite.client
-import snakebite.errors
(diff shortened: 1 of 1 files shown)


matz-e:
Great! Thank you! 🎉

PR #610, opened 2017-10-25
Allow adjustable tasksizes for prod workflows

annawoodard:
Before, users could specify `events_per_task`, `events_per_lumi` (optionally) and `number_of_tasks`. The tasksize adjusts to meet the requested runtime, but has no effect on the number of events processed per task. This changes the dataset specification to `total_events` (what the user ultimately cares about), `events_per_lumi`, and `lumis_per_task` and adjusts `events_per_task` according to the tasksize. In this way, adjustable tasksizes have the same effect as for non-prod workflows.

  workflows)

### DIFF
--- a/examples/mc_gen.py
+++ b/examples/mc_gen.py
@@ -25,9 +25,9 @@ lhe = Workflow(
     sandbox=cmssw.Sandbox(release='mc_gen/CMSSW_7_1_16_patch1'),
     merge_size='125M',
     dataset=ProductionDataset(
-        events_per_task=250,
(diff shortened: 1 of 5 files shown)


matz-e:
Makes sense? I think for the standard way of producing MC, we'd have to add the requirement to only allow to draw units from the database with consecutive luminosity sections without gaps, as to not introduce doubled lumi sections?

That being said, thank you for laying the ground-work for that! 👏 🎉 🌮 I don't think much should change for conventional MC production à la ttH.

PR #613, opened 2017-10-25
Fix silent crashes

annawoodard:
This fixes two error modes which were causing silent crashes when not in foreground mode. For cases when the user specifies non-existent `extra_inputs`, it is not straightforward to get that check to happen before making the working directory without substantial reshuffling, so I think it makes sense to exit right away anyways instead of raising. Also: miscellaneous minor tweaks.

  workflows)


matz-e:
Thank you! 👏 🎉

PR #618, opened 2017-12-12
Make commands a lot more customizable.

### LINKED ISSUE #207: Custom commands do not get input files passed

matz-e:
Remove `pset` and `argument` parameters, and instead parse the `command`
parameter for workflows. This allows to shuffle things around a little and
will be more flexible in the future. The following merge command would now
be possible:

    my_merge.py -o @outputfiles --inputs @inputfiles

While at it, pulling in datasets from DBS and splitting them on files broke
recently, fix it.

Fixes #207.

  workflows)


lobster/core/data/task.py
```diff
@@ -400,8 +423,12 @@ def copy_inputs(data, config, env):
         # When the config specifies no "input," this implies to use
         # AAA to access data in, e.g., DBS
         if len(config['input']) == 0:
-            config['mask']['files'].append(file)
-            config['file map'][file] = file
+            if config['executable'] == 'cmsRun':
+                filename = file
+            else:
+                filename = "root://cmsxrootd.fnal.gov/" + file
```
annawoodard: I'm always wary of adding more knobs until we know we need them, but maybe should this be configurable?

lobster/core/workflow.py
```diff
@@ -151,11 +151,13 @@ class Workflow(Configurable):
             :class:`ValueError` will be raised otherwise.
         command : str
             Which executable to run (for non-CMSSW workflows)
```
annawoodard: `(for non-CMSSW workflows)` should be removed now, no?

lobster/core/workflow.py
```diff
@@ -151,11 +151,13 @@ class Workflow(Configurable):
...
+
+            The executable string may contain `@args`, `@outputfiles`, and
+            `@inputfiles`, which will be replaced by unique arguments and
+            output as well as input files, respectively.
```
annawoodard: Consider clarifying here what the default is (i.e. so it is obvious to use `cmsRun simple_pset.py` instead of `cmsRun simple_pset.py @inputfiles`).

matz-e:
Thank you for the review! I hope that I've addressed your comments sufficiently. I've taken the opportunity to change the task wrapper to be able to write tests for it, too.

lobster/core/data/task.py
```diff
@@ -183,6 +167,20 @@ def replace(xs, s, ys):
     return newcmd
 
 
+def find_xrootd_server(filename):
+    """Find the leading XRootD server in `filename` and return it.
+    """
+    fakepath = '/store/user/foo/bar.root'
+    doc = xml.dom.minidom.parse(filename)
+    for e in doc.getElementsByTagName("lfn-to-pfn"):
+        if e.attributes["protocol"].value != "xrootd":
```
annawoodard: Since there might be fallback xrootd servers configured, we could check if `xrootd in e.attributes["protocol"].value` instead, and cycle through them in case the first one failed.

This is ultra-low priority-- I'm a big believer in not adding additional complexity unless there's a clear need/use case. So I'll open a separate issue just to document we considered it, and if the issue actually ever comes up, we could implement it.

annawoodard:
This is awesome. Thanks @matz-e! 👏 👏 👏

PR #621, opened 2017-12-14
Allow to filter on file extensions.

matz-e:

  workflows)

### DIFF
--- a/lobster/core/dataset.py
+++ b/lobster/core/dataset.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+import fnmatch
 import math
 import os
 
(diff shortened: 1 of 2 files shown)


lobster/core/dataset.py
```diff
@@ -67,23 +84,28 @@ class Dataset(Configurable):
             A list of files or directories to process.  May also be a `str`
             pointing to a single file or directory.
         files_per_task : int
-            How many files to process in one task
+            How many files to process in one task. Defaults to 1.
```
annawoodard: Is it necessary to say the default explicitly, since it's specified in the constructor?

matz-e: I think for the documentation, this is nice? It's right there by the description and you don't have to hunt for it. We also specify the default in a bunch of other docstrings.

annawoodard:
Thanks @matz-e!!

PR #627, opened 2018-03-10
Fix output errors

### LINKED ISSUE #626: Stageout Sometimes Treated as Failure Because of Problem with Remote File Size Check
I've noticed on the dedicated T3 workers, sometimes stageout is marked as a failure because it tries to use a local operation (i.e. `stat`) to measure the size of the remote file, but the size is returned as "0".

I'm having trouble believing that the size is really zero because I see a consistent pattern of attempt to do a local file copy, measure file size zero, attempt to do a remote copy, and fail because file already exists.  I modified stageout with `gfal-copy` to use `-f` so that it will overwrite files that exist, and it still failed because `check_output` uses all the methods defined in the config to access the output, so if `file:` is before `gsiftp:` then even if it's `gsiftp:` that's used to copy, it will still use `stat` to check the size if it can and fail with size = 0.  The problem isn't consistent in that some jobs fail and some succeed.  I suspect it's an issue with FUSE on some of the hosts.

The solution I'm contemplating is to treat mismatched file sizes as a failure of the file size measurement rather than as a failure of the transfer.  So if the local and remote files don't match, Lobster would try to measure the file size with the next available method.  File transfers would only be counted as failures if all methods of measuring file size failed.

There is one other potential problem here.  `check_output` is coded so that if none of the implemented file size check methods succeed, then it returns `True`.  The comments suggest this is a compromise because there are transfer methods (e.g. `gsiftp:`) which don't have an implemented file measurement technique, so it's possible to construct a config where outputs could be transferred but not have their size checked.  To get around this, I am think I'll implement file size checks using `gfal-stat` and `hadoop dfs -stat "Size: %b"` to close the loopholes where you could transfer but not check.  (I'm also planning to implement hdfs stageout using `hadoop dfs -put`.)  Then I don't think there will be any reason to pass through cases where we didn't manage to measure the remote file size, so if no method succeeds, that will be treated as a failure (e.g. `check_output` returns `False`).

I know you have both moved on, but @matz-e or @annawoodard, if you have a moment to look over this plan and see any flaws with it, please let me know.

klannon:
Fixes #626.  Also fixes a bug I noticed when trying to diagnose #626 where edm files were treated as non-edm files because of a typo in the list of branches expected in edm files.  Also fixes an issue noticed while testing the fix to #626 where WQ gets confused about `report.json` which is both an input and output file.  See 5b37cba.  I changed the cctools version here to the one I used to test the fixes, but none of this new code *requires* a new cctools version.  Therefore, I'm not updating the Lobster version number in this case.

  workflows)


matz-e:
Looks superb!

PR #635, opened 2018-07-18
Better determination of output files.

### LINKED ISSUE #317: Need to determine outputs with the right CMSSW setup
Right now, when we use Lobster with, e.g., `CMSSW_7_4_1`, but have not all output files specified in a config using a separate sandbox release, we may have an exception when loading a pset for, e.g., `CMSSW_8_0_X`, or for a release with a lot of custom packages, because python imports are missing.

We should create a little script that is executed within the sandbox environment and determines the output files there, and passes this back to Lobster (i.e. via subprocess). I think this relates to the discussion in #314.

### LINKED ISSUE #634: Auto-Detection Fails when Multiple Releases Used in Same Lobster Project
As reported by @dmf1998, Lobster doesn't do the right thing if you have multiple workflows in one project that use different CMSSW releases.  In order to auto-detect the output or global tags used in the config of each workflow, Lobster attempts to load the pset, but it does it with whichever CMSSW release was used to start Lobster (e.g. where `cmsenv` was issued) instead of the one that the workflow needs.  If the `cmsRun` config is not compatible with the release used to start Lobster, the auto-detection of global tag or output file fails and Lobster terminates.  Fixing this bug is not straightforward, as Lobster would need to interpret each workflows pset in the appropriate release.  Workaround is just to specify outputs and global tags explicitly.  Because global tags are only needed for publishing, if you don't plan to publish, you can set `globaltag = 'dummy'` to avoid triggering global-tag auto-detection.

matz-e:
A rough draft for a script to automatically determine outputs and
global tags from psets that require a release other than the current
active one. Somehow #634 triggered me to do this, since this has been one of my major pet-peeves all along. Since I'm currently not able to test this, additional work may be required, so maybe this is just a dead end? Feel free to built on top or discard!

This is also just a quick hack; Ideally, there would be a dedicated `cmssw.workflow` module/class that takes care of this in a cleaner way.

Maybe fixes #317 and maybe fixes #634. Additional changes may have to
be applied.

  workflows)

### DIFF
--- a/lobster/core/data/autosense.sh
+++ b/lobster/core/data/autosense.sh
@@ -5,7 +5,7 @@ pset=$2
 shift
 shift
 
-if [[ $# -ge 1 || -n "$release" || -n "$pset" ]]; then
+if [[ -z "$release" || -z "$pset" ]]; then
 	echo "usage: autosense.sh release pset args..."
 	exit 1
 fi
@@ -13,6 +13,7 @@ fi
 source /cvmfs/cms.cern.ch/cmsset_default.sh 
 cd "$release"
 eval $(scramv1 runtime -sh)
+cd - > /dev/null
 
 python <<EOF
 import imp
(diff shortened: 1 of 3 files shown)


klannon:
Thanks for putting this together. I can take a look at this when I’m back from vacation.

klannon:
Hey @matz-e!  So after too long a time of ignoring this PR, I actually got around to testing it.  There were a number of bug fixes I had to add to make it work.  I'd like to commit my fixes and then merge this PR, but I don't seem to have permission to push to your repo any more.  Alternatively, I could just push my version to a new branch in the NDCMS lobster repo, close this PR, and carry on from there.  However, I'd prefer to do it through this PR since you're the one who got it started.  Any suggestions?

matz-e:
Hi @klannon, that's great! I re-forked this repo after we moved it into NDCMS, I've just re-added you as a collaborator on my fork. Although according to the little checkbox on the PR, you should be able to push to my `autosense` branch if you're a Lobster maintainer…

If you have any questions about the PR, I can try to recollect my thoughts about it and will answer them as best as possible!

klannon:
I've been waiting to do more testing on this, but it fixes an important problem that people sometimes run into, so I'm just going to go ahead and merge it and we'll fix the problems that haven't been caught yet when they're reported.

PR #640, opened 2020-01-07
Update setup.py

khurtado:
Changes:

- [ ] Are all additional dependencies in `setup.py`?
Update `setup.py`  with new snakebite version to handle hadoop 2.6.


klannon:
So, our CI is broken, but this is a simple PR, so I'm just going to merge it...

PR #644, opened 2020-04-26
Attempt to fix travis by installing TLS headers.

matz-e:
I'm noticing this in my Travis feed, and it seems like an easy fix.


klannon:
Sorry for leaving this languish.  Since this is for Travis, I might as well merge it.  It won't break anything else and Travis is already broken.  😞

PR #693, opened 2025-01-29
Changes for el9-based cmssw versions

anpicci:
I have slightly modified the scripts in order to properly handle the CMSSW sandboxes that require el9 architecture


btovar:
@anpicci Note that with this branch we are trying to move away from parrot. The images from runos should have given you cvmfs, isn't that the case?

Andrew42:
@anpicci running with your changes I came across a crash caused by a byte-string getting introduced to data which then gets serialized into a json output file. Specifically, I had to change [this](https://github.com/NDCMS/lobster/blob/lobster-python3/lobster/core/data/task.py#L207) line to
```python
checksum = stdout.split()[-2].decode('utf-8')
```

Andrew42:
In addition to the byte-string bug, I also had to modify [task.py LL71-73](https://github.com/NDCMS/lobster/blob/lobster-python3/lobster/core/data/task.py#L71-L73) to exclude this check for `CMSSW_12_4_X`. The `XrdAdaptor::XrdStatisticsService` module appears to not exist in this CMSSW release. I simply removed/commented the `process.add_(xrdstats)` to keep it from being added to the CMS executable.

Without this change the lobster task will make it through set up, but crash once it attempts to do the `cmsRun` executable of the form:
```
>>> executing 'cmsRun -j report.xml test_LHE_cfg_mod.py' @ Fri Oct  3 17:04:24 2025 
>> using /tmp/tmpmgxbm8kh to store command output
>> cmd: ----- Begin Fatal Exception 03-Oct-2025 17:04:28 EDT-----------------------
>> cmd: An exception of category 'PluginNotFound' occurred while
>> cmd: [0] Constructing the EventProcessor
>> cmd: Exception Message:
>> cmd: Unable to find plugin 'XrdAdaptor::XrdStatisticsService' in category 'CMS EDM Framework Service'. Please check spelling of name.
>> cmd: ----- End Fatal Exception -------------------------------------------------
```
Removing the module did not seem to cause any other issues for the task, though likely results in some missing monitoring metrics.

Andrew42:
@anpicci is [this](https://github.com/NDCMS/lobster/pull/693/files#diff-59c10ea06adfa9a612c9b38210cce86171e48d8482818b660d70ec26ac88c1baR72) change missing an 'and'?

Andrew42:
@anpicci another thing that I needed to change was the `ndcctools` version that gets installed in the `conda` environment.

Right now the [yaml file](https://github.com/NDCMS/lobster/blob/lobster-python3/lobster_env.yaml#L73) has it pinned to `7.11.1`, however, in order to use the `--runos al8-wq-7.15.9` option for the WQ factory, @btovar said that I should use `7.15.9` or newer. The version of `ndcctools` I updated to in my working environment was `7.15.11`, so we can either use that or pin it to `7.15.9`.

